PR #57, opened 2024-10-02
Add build parameters in detailed component

adellinekania:
Extend `/rest/api/2/components/{component}/versions/{version}` end-point with build related data

### DIFF
--- a/component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
+++ b/component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
@@ -99,6 +99,7 @@ class EscrowConfigValidator {
                 validateClientCode(moduleConfig, componentName)
                 validateReleasesInDefaultBranch(moduleConfig, componentName)
                 validateSolution(moduleConfig, componentName)
+                validateBuildConfigurationTools(moduleConfig)
             }
         }
         if (!hasErrors()) {
@@ -408,6 +409,25 @@ class EscrowConfigValidator {
         def solution = moduleConfig.getSolution()
         if (solution == null) {
             registerError("solution is not specified in '$component'")
+        }
+    }
+
+    def validateBuildConfigurationTools(EscrowModuleConfig moduleConfig) {
+        def tools = moduleConfig.getBuildConfiguration()?.getTools()
+        tools?.each { tool ->
+            def toolName = tool.getName()
+            if (toolName== null)  {
+                registerError("tool is not configured correctly")
+            }
+            if (tool.getEscrowEnvironmentVariable() == null) {
+                registerError("tool escrowEnvironmentVariable is not specified in '$toolName'")
+            }
+            if (tool.getSourceLocation() == null) {
+                registerError("tool sourceLocation is not specified in '$toolName'")
+            }
+            if (tool.getTargetLocation() == null) {
+                registerError("tool targetLocation is not specified in '$toolName'")
+            }
         }
     }
 
--- a/component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
+++ b/component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
@@ -417,7 +417,7 @@ class EscrowConfigValidator {
         tools?.each { tool ->
             def toolName = tool.getName()
             if (toolName == null)  {
-                registerError("tool is not configured correctly")
+                registerError("tool name is not specified")
             }
             if (tool.getEscrowEnvironmentVariable() == null) {
                 registerError("tool escrowEnvironmentVariable is not specified in '$toolName'")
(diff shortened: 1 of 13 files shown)


pgorbachev:
@adellinekania , 
1) could please add tests for the changed functionality
2) add an example of the new section to the Description

pgorbachev (changes requested):
Tests should be added

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/ComponentsRegistryServiceControllerTest.kt
```diff
@@ -393,6 +395,53 @@ class ComponentsRegistryServiceControllerTest : BaseComponentsRegistryServiceTes
...
+                    ToolDTO(
+                        name = "PowerBuilderCompiler170",
+                        escrowEnvironmentVariable = "PBC_BIN",
+                        sourceLocation = "\$env.PBC/170",
+                        targetLocation = "tools/auto_compiler"
+                    )
+                ),
+            )
+        }
+        Assertions.assertEquals(expectedComponent, actualComponent)
```
pgorbachev, line 442: Please check only buildParameters DTO. Let's make test isolated from other parameters

test-common/src/test/resources/components-registry/common/TestComponents.groovy
```diff
@@ -840,3 +840,16 @@ TEST_COMPONENT4 {
         tag = 'tdsecure-$version'
     }
 }
+
+"TEST_COMPONENT_BUILD_PARAMETERS" {
```
pgorbachev, line 844: COMPONENT_WITH_BUILD_PARAMETERS

pgorbachev (approved):

ozonophore (approved):

components-registry-service-core/src/main/kotlin/org/octopusden/octopus/components/registry/core/dto/ToolDTO.kt
```diff
@@ -0,0 +1,14 @@
+package org.octopusden.octopus.components.registry.core.dto
+
+import com.fasterxml.jackson.annotation.JsonCreator
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ToolDTO @JsonCreator constructor(
+    @JsonProperty("name") val name: String? = null,
```
aryabokon, line 9: can we make this properties as non-nulable?

adellinekania: Yes, it is possible to make the name property non-nullable, as the corresponding `Tool` data class in the `component-resolver-api` already defines it as non-nullable. However, I would need confirmation on whether every Tool instance is guaranteed to have a name. If so, I can safely enforce non-nullability here.

aryabokon: @adellinekania , you need to have guarantee via validation on EscrowConfig loading

@pgorbachev , what about another fields? Do we have a contract?

pgorbachev: @adellinekania , 
All the parameters are mandatory except install script.
So 
1) Add validateBuildParameters to EscrowConfigValidator.groovy to ensure this (add test for this in EscrowConfigurationLoaderTest)
2) make DTO fields non-null except install script

component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
```diff
@@ -411,6 +412,25 @@ class EscrowConfigValidator {
         }
     }
 
+    def validateBuildConfigurationTools(EscrowModuleConfig moduleConfig) {
+        def tools = moduleConfig.getBuildConfiguration()?.getTools()
+        tools?.each { tool ->
+            def toolName = tool.getName()
+            if (toolName== null)  {
```
pgorbachev, line 419: Please, reformat the change code

pgorbachev (dismissed):

component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/EscrowConfigValidator.groovy
```diff
@@ -411,6 +412,25 @@ class EscrowConfigValidator {
         }
     }
 
+    def validateBuildConfigurationTools(EscrowModuleConfig moduleConfig) {
+        def tools = moduleConfig.getBuildConfiguration()?.getTools()
+        tools?.each { tool ->
+            def toolName = tool.getName()
+            if (toolName == null)  {
+                registerError("tool is not configured correctly")
```
CamaradeRoman, line 420: let's change error description to more specific, like "tool name is not specified"

pgorbachev (approved):

CamaradeRoman (approved):

components-registry-service-core/src/main/kotlin/org/octopusden/octopus/components/registry/core/dto/ToolDTO.kt
aryabokon: @adellinekania , we talked about nonNullable fields and fields still nullable

aryabokon (changes requested):
deleted

Issue #60, opened 2024-10-18
Validate distribution->docker parameter


pgorbachev:
Check that distribution->docker matches  "dockerImageName:tag[,dockerImageName2:tag2]*" format
 (in terms of https://docs.docker.com/reference/cli/docker/image/tag/), 
SpEL is not allowed in docker image name. SpEL is supported only in tag.

Add validation for "distribution->docker" field. 
Example:
docker = 'abc/abc-mdes-client:${version},abc/abc-mdes-listener:${version},abc/abc-mdes-emulator:${version}-jdk11'

see https://github.com/octopusden/octopus-components-registry-service/blob/f5d82fc68b44060ab65dddc596b4c07dca7511f3/component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/GroovySlurperConfigValidator.groovy#L242

NOTE: The Task is required for calculation of component by docker image

pgorbachev:
+ Update documentation for distribution->docker

PR #75, opened 2025-04-07
validate distribution/docker field

fishinitself:
see #60 
dynamic context variables in expressions, like ${env.JAVA_HOME} or ${baseDir} are prohibited.
version-related variables, like ${version} or ${major} are allowed


component-resolver-core/src/test/groovy/org/octopusden/octopus/escrow/configuration/validation/GroovySlurperConfigValidatorTest.groovy
```diff
@@ -39,4 +41,32 @@ class GroovySlurperConfigValidatorTest extends GroovyTestCase {
         assert !DOCKER_PATTERN.matcher("org.octopusden/octopus:image:1.0").matches()
         assert !DOCKER_PATTERN.matcher("org.octopusden/octopus/image:.0").matches()
     }
+
+
+    void testDockerField() {
```
adellinekania, line 46: I think it would be better to give the test a less general name—something like `testDockerNoEnvVarExpr`—so it’s clearer that the test checks for disallowed environment variable expressions.

pgorbachev:
Negotiated with community:
CRS: фича вычисления component name by docker image. Договорились сделать валидацию, что в distribution->docker в image name нет SPEL + проверка на уникальность.

Issue #87, opened 2025-05-28
Matching type of version by string version


fishinitself:
`normalizeVersion` in `EscrowConfigurationLoader` looks for match with type of version by string version. 
To work properly the search should be performed from more many-component (more specific) types to less many-component (more general) types, and hot fix should be the first. Currently it is the last, so hot fix version can't correctly matched. The correct order is 
- Hot fix 
- Build 
- Release 
- Major 
- Line

pgorbachev:
Resolved. `normalizeVersion` now tries the hotfix format **first** when hotfix is enabled (ordering Hotfix → Build → Release → Major → Line), in `JiraComponentVersionFormatter` (hotfix-enabled branch checked before Build). Fixed by commit 13d6102 (2025-07-17). Closing as completed.

PR #94, opened 2025-06-17
Matching type of version and formatting #87

fishinitself:


component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/loader/EscrowConfigurationLoader.groovy
```diff
@@ -135,8 +135,14 @@ class EscrowConfigurationLoader {
         def jiraComponentVersionFormatter = new JiraComponentVersionFormatter(versionNames)
         def numericVersion = new NumericVersionFactory(versionNames).create(version)
 
-        def formats = [
-                //TODO: [jiraComponentVersionFormatter.&matchesHotfixVersionFormat, jiraComponentVersionFormatter.hotfixVersionFormat],
+        def isHotfixEnabled = vcsSettings?.getVersionControlSystemRoots()?.any { it.hotfixBranch?.isEmpty() == false }
```
pgorbachev, line 138: We might need this logic in one place in the project (that determine if HF is enabled for the component)

pgorbachev (changes requested):

component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/loader/LoaderHelper.groovy
```diff
@@ -0,0 +1,10 @@
+package org.octopusden.octopus.escrow.configuration.loader
+
+
+import org.octopusden.octopus.escrow.model.VCSSettings
+
+class LoaderHelper {
+    static boolean isHotFixEnabled(VCSSettings vcsSettings) {
```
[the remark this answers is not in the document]
pgorbachev: It should be used also in EscrowConfigValidator#validateHotfixVersionFormat

component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/loader/ComponentHotfixSupportResolver.groovy
```diff
@@ -3,8 +3,8 @@ package org.octopusden.octopus.escrow.configuration.loader
 
 import org.octopusden.octopus.escrow.model.VCSSettings
 
-class LoaderHelper {
-    static boolean isHotFixEnabled(VCSSettings vcsSettings) {
+class ComponentHotfixSupportResolver {
+    boolean isHotFixEnabled(VCSSettings vcsSettings) {
```
pgorbachev, line 7: I would suggest passing ModuleConfig instead of vcsSettings for the sake of flexibility

fishinitself: That would be great but we cant
this is my previous comment 

_org.octopusden.octopus.escrow.model.VCSSettings is the most common parameter - we have invocation context with JiraVersionRange and with EscrowModuleConfig. Probably we can extract common for JiraVersionRange and EscrowModuleConfig interface and use it but I don't like this._

in other words call we have no EscrowModuleConfig here - 
components-registry-service-[server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentRegistryResolverImpl.kt](https://github.com/octopusden/octopus-components-registry-service/blob/49c0772bb8e320644a7ea634afad4e5a006f9118/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentRegistryResolverImpl.kt#L383)

pgorbachev (dismissed):

CamaradeRoman (changes requested):

PR #171, opened 2026-05-05
Remove QueryMap annotation from getAllComponents

rodemark:
`systems` is a regular query parameter list, but it was annotated with `@QueryMap.`
This worked on JDK 8, but on JDK 21 Feign tries to encode the list via reflection and fails with `InaccessibleObjectException`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parameter handling in the components registry service client for improved query processing.


qu0d (approved):

qu0d (approved):

pgorbachev (approved):
Looks good to me. I verified the Feign client change locally; the relevant client test and :components-registry-service-client:check pass.

PR #199, opened 2026-05-15
fix(import): preserve Defaults.requiredTools on schema-v2 import

pgorbachev:
## Summary
- Components that don't declare own `build.requiredTools` but inherit `BuildEnv` from `Defaults.groovy` returned empty `buildParameters.tools` on the schema-v2 DB candidate (Git baseline kept `[BuildEnv]`).
- Two root causes (both fixed): (a) the Groovy loader's per-range merge drops Defaults-inherited `requiredTools` when the component declares its own top-level `build { ... }` block; (b) the import previously wrote a `BUILD_REQUIRED_TOOLS` marker at `savedBase.versionRange`, which for synthetic-base / multi-range components is only the DSL's first range — version queries against any other range silently bypassed it.
- Fix: attach base required-tools junctions directly to the BASE row (reader already falls back to `base.requiredToolJunctions`); fall back to `Defaults`' tools when the per-component merge came back empty; align `emitMarkerOverrides` "tools differ?" comparison to use the same effective base.

## Regression test
- `GitVsDbValidationTest.VAL-011 defaults requiredTools git vs db parity` — switches each candidate between Git and DB sources, compares `buildParameters.tools`, fails on any divergence.
- Reproduced on new synthetic fixture `legacy-multi-range-tool-inherit` (multi-range, top-level `build { javaVersion }`, no own `requiredTools`). Pre-fix: `git=[BuildEnv] db=[]`. Post-fix: parity on both sides of the `1.0.107` range boundary. Existing single-range / no-top-level-build components keep passing.

## Acknowledged limitation
The Defaults fallback can't distinguish "loader merge dropped tools" from "component explicitly cleared tools" — both surface as an empty list at the same call site. No current DSL component exercises the latter; preserving that distinction would need loader changes (kdoc note inline).

## Test plan
- [x] `AUTH_SERVER_URL=http://localhost:9999 AUTH_SERVER_REALM=octopus ./gradlew build -x dockerBuildImage -x dockerPushImage -x ocCreate -x ocProcess -x :components-registry-automation:test` exits 0
- [x] VAL-011 reproduced pre-fix, green post-fix on both probe versions
- [x] Independent Sonnet subagent review pre-push; both P2 findings (redundant marker on per-range tool overrides + extra range-boundary probe) addressed in the same commit
- [ ] CI green

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ImportServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ImportServiceImpl.kt
@@ -502,6 +502,28 @@ class ImportServiceImpl(
         }
     }
 
+    /**
+     * Common-defaults tools (`Defaults.groovy` → `build { requiredTools = "..." }`),
+     * lazy-loaded once per `ImportServiceImpl` instance and reused as a fallback
+     * when a component's own merged `buildConfiguration.tools` came back empty
+     * (the Groovy loader drops Defaults-inherited tools whenever the component
+     * declares its own `build { ... }` block — see `importModule` for context).
+     *
+     * Limitation: this fallback cannot distinguish "loader merge dropped tools"
+     * from "component explicitly cleared tools" — both surface as an empty list
+     * on `EscrowModuleConfig.buildConfiguration.tools`. A future opt-out for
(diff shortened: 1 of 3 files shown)


pgorbachev:
Addressed both Copilot inline comments in b3ef818:
- L500 kdoc reworded to reflect the hand-picked candidate list (no auto-discovery).
- L537 `System.err.println` removed; the failure-side `failures` list already carries the diff.

PR #202, opened 2026-05-15
feat(api): domain-named meta endpoints for option lists

pgorbachev:
## Summary

Schema-v2 left three aspect string fields free-form on the wire — `BuildAspect.buildSystem`, `VcsEntry.repositoryType`, `Escrow.generation` are now `string`, not enum. The canonical valid-token sets still live as Kotlin enums but are no longer reachable through the v4 typed contract, so admin and portal users have no way to discover them.

This PR adds three GET endpoints under `/rest/api/4/components/meta/*` that return the option lists:

| Endpoint | Source |
|---|---|
| `GET /meta/build-systems` | `BuildSystem.values()` (9 values: GRADLE, MAVEN, WHISKEY, …) |
| `GET /meta/repository-types` | `RepositoryType.values()` (3: GIT, MERCURIAL, CVS) |
| `GET /meta/escrow-generations` | `EscrowGenerationMode.values()` (3: AUTO, MANUAL, UNSUPPORTED) |

Names are domain-named, NOT implementation-named (no `/meta/enums`), so the wire surface survives a future move of the option source from a Kotlin enum to a config table or admin-editable registry.

All three require `ACCESS_COMPONENTS` via `@PreAuthorize`, matching the existing `/meta/owners` pattern.

## Portal context

Portal-side wiring already landed on the schema-v2 feature branch (`feature/crs-schema-v2`, commit `6ad8876`): a new `useFieldOptions` hook calls these endpoints when the admin field-config registry has no explicit `options[]` seeded. Until this PR merges, the hook 404s gracefully and `EnumSelect` collapses to a degenerate "None + current value" dropdown — users on a fresh CRS install cannot change Build System on a component once it has a value set.

Cross-repo TD record: portal `docs/tech-debt/TD-005-schema-v2-followups.md` item #10.

## History

Originally opened as #201 stacked on PR #193 (`feat/schema-v2-phase5b-6`). After #193 merged into `feat/schema-v2-sql` the base branch was deleted and #201 auto-closed. This is a clean re-open against `feat/schema-v2-sql` (PR #192), with the single endpoint commit rebased on the new base.

## Test plan

- [x] `MetaOptionsEndpointsTest` — three cases asserting each endpoint returns the canonical enum value set in declaration order, under the `ft-db` profile
- [x] Full `:components-registry-service-server:test` suite green
- [ ] Manual smoke once portal feature branch is rebased: open `/components/{id}` Build tab, confirm dropdown lists all 9 build systems

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/ComponentControllerV4.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/ComponentControllerV4.kt
@@ -1,9 +1,9 @@
 package org.octopusden.octopus.components.registry.server.controller
 
-import org.octopusden.octopus.components.registry.core.dto.BuildSystem
-import org.octopusden.octopus.components.registry.core.dto.EscrowGenerationMode
-import org.octopusden.octopus.components.registry.core.dto.RepositoryType
+import org.octopusden.octopus.components.registry.api.enums.EscrowGenerationMode
 import org.octopusden.octopus.components.registry.core.exceptions.NotFoundException
+import org.octopusden.octopus.escrow.BuildSystem
+import org.octopusden.octopus.escrow.RepositoryType
 import org.octopusden.octopus.components.registry.server.dto.v4.ComponentCreateRequest
 import org.octopusden.octopus.components.registry.server.dto.v4.ComponentDetailResponse
 import org.octopusden.octopus.components.registry.server.dto.v4.ComponentFilter
@@ -55,10 +55,17 @@ class ComponentControllerV4(
     // Domain-named option lists for the three free-form aspect string fields
     // (buildSystem, repositoryType, generation). The portal's EnumSelect uses
     // these to populate dropdowns when the admin field-config registry has no
-    // explicit options[] seeded. The endpoint names are domain-named, not
+    // explicit options[] seeded. Endpoint names are domain-named, not
     // implementation-named (no `/meta/enums`), so the wire surface survives
     // a future move of the option source from a Kotlin enum to a config
     // table or admin-editable registry.
+    //
+    // The enums sourced here are the *persistence-layer* ones used by
+    // `EntityMappers` on read/write — NOT the `core.dto.*` mirrors. The DTO
+    // variant of `BuildSystem` carries `NOT_SUPPORTED` while
+    // `EntityMappers.safeParseBuildSystem` calls `BuildSystem.valueOf` on
+    // the escrow variant (which has `ESCROW_NOT_SUPPORTED`). Advertising the
+    // DTO token would silently drop that value on save.
     @GetMapping("/meta/build-systems")
     @PreAuthorize("@permissionEvaluator.hasPermission('ACCESS_COMPONENTS')")
     fun getBuildSystems(): List<String> = BuildSystem.values().map { it.name }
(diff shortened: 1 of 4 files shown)


pgorbachev:
Addressed both Copilot inline review findings from the originally-opened #201 (auto-closed when its `feat/schema-v2-phase5b-6` base was deleted on PR #193 merge).

Both comments pointed at the same root issue: there are two `BuildSystem` enums in the codebase, and the controller + test were sourcing from the DTO mirror (`core.dto.BuildSystem`, token `NOT_SUPPORTED`) while `EntityMappers.safeParseBuildSystem` parses against the persistence variant (`org.octopusden.octopus.escrow.BuildSystem`, token `ESCROW_NOT_SUPPORTED`). Advertising the DTO token would have silently dropped that selection on save.

Fix in `aed1b00`: moved all three endpoints to the persistence-layer enums (build-systems, repository-types, escrow-generations) so the endpoint advertises exactly the values that round-trip through `EntityMappers`. The test now asserts against the persistence enums and adds a dedicated regression-lock case (`buildSystems_advertisesPersistenceTokens`) that explicitly checks `ESCROW_NOT_SUPPORTED` is present and the DTO-only `NOT_SUPPORTED` is absent — so a future revert to the wrong source surface trips immediately.

Full `:components-registry-service-server:test` suite green (2m 16s).

pgorbachev:
Addressed all three Copilot inline findings.

**Findings #1 and #2** (controller + test sourcing from `core.dto.BuildSystem` instead of `escrow.BuildSystem`) — already fixed in \`aed1b00\` before the review batch posted; Copilot reviewed the rebased \`a70ad7c\` and the comments rendered after my push of the fix. The persistence enums are now the source for all three endpoints (\`escrow.BuildSystem\`, \`escrow.RepositoryType\`, \`api.enums.EscrowGenerationMode\`), and the regression-lock test \`SYS-038 GET meta build-systems uses persistence enum tokens not DTO mirror\` explicitly checks for \`ESCROW_NOT_SUPPORTED\` presence and \`NOT_SUPPORTED\` absence.

**Finding #3** (AGENTS.md §Test-to-Requirement Traceability — new feature tests must be tied to a numbered requirement, and the ID must appear in both method name and \`@DisplayName\`) — fixed in \`6332f7e\`:

- \`docs/db-migration/requirements-common.md\`: added \`SYS-038: Domain-named meta endpoints for free-form aspect option lists\` with motivation, four acceptance criteria, and the persistence-vs-DTO enum-source decision recorded.
- \`MetaOptionsEndpointsTest\`: every test method renamed to a backticked \`SYS-038 …\` form; every \`@DisplayName\` prefixed with \`SYS-038:\`. Pattern mirrors \`ListComponentsOwnerFilterTest\` (SYS-035).

\`:components-registry-service-server:test\` suite green.

pgorbachev:
Addressed all 7 inline findings from the 18:13Z review batch in \`187967e\`:

1. **ktlint import order — ComponentControllerV4** — moved \`escrow.*\` imports after \`components.registry.server.*\`.
2. **ktlint import order — MetaOptionsEndpointsTest** — moved \`server.support.viewerJwt\` before \`escrow.*\`.
3. **SYS-038 precondition wording** — replaced "Caller authenticated with ACCESS_COMPONENTS" with "Caller has ACCESS_COMPONENTS (granted to ROLE_ANONYMOUS by default — no JWT required)" to match the v4 read posture.
4. **SYS-038 description** — narrowed: "persistence for buildSystem/repositoryType; API enum for generation" with the rationale that \`Mappers.toDTO()\` reads \`api.enums.EscrowGenerationMode\` directly off the escrow model.
5. **Test class-doc** — same narrowing applied to the comment block over MetaOptionsEndpointsTest.
6. **Controller comment** — same narrowing applied to the comment over the three meta-* endpoints.
7. **implementation-progress.md** — added a Phase 9 row for "Meta option-list endpoints" referencing PR #202 and SYS-038, per the AGENTS.md "update implementation-progress after each completed task" rule.

Focused test (\`MetaOptionsEndpointsTest\`) green locally (40s).

PR #203, opened 2026-05-15
fix(v4): enum + range syntax validation on write-side (close PR #192 review)

pgorbachev:
## Summary
Closes the 6 remaining "STILL VALID" Copilot review findings on PR #192 + 1 weak-test finding. Pre-fix the v4 API accepted enum-typed fields and \`versionRange\` values without runtime validation; the resolver later parsed those strings with \`Enum.valueOf\` and silently dropped the row or returned null on read, surfacing as a 5xx.

**Write-side guards added:**
- \`productType\` (CREATE + PATCH) — vs \`ProductTypes.values()\`.
- \`baseConfiguration.build.buildSystem\` (CREATE + PATCH) — vs \`BuildSystem.values()\`.
- \`vcsEntries[].repositoryType\` — inside \`replaceVcsEntries\` so both base-config AND field-override \`vcs.settings\` marker payloads are covered (one guard, two paths).
- \`packages[].packageType\` — inside \`replacePackages\` (same dual-path coverage), restricted to the resolver-supported \`{"DEB","RPM"}\` set.
- \`build.buildSystem\` SCALAR_OVERRIDE — \`ConfigurationRowAccessors.applyScalarValue\` routes through new \`requireBuildSystem\`.
- \`baseConfiguration.versionRange\` — pre-existing validation; PR adds non-regression test.

**Test strengthening:** \`ListComponentsSystemFilterTest\` now verifies the v2 junction-based \`?system=\` filter actually filters (CLASSIC returns rows with CLASSIC; unknown system returns empty), not just that the endpoint returns 200.

**Regression:** new \`V4WriteValidationTest\` — 8 tests covering all write paths above. All fail pre-fix (some 201, one 409); all green post-fix.

## Independent Sonnet review (pre-commit)
- **P1 fixed:** field-override marker payloads were bypassing the new enum guards. Moved validation into \`replaceVcsEntries\` / \`replacePackages\` so both base-config AND \`applyMarkerChildren\` paths are covered.
- **P2 fixed:** missing \`@DirtiesContext\` on \`V4WriteValidationTest\` (seeds real components → context-cached re-run would 409).
- **P2 fixed:** redundant \`validateRangeSyntax\` call in updateComponent (pre-existing validation reused, no duplicate).
- **P2 noted:** \`BUILD_SYSTEM_NAMES\` duplicated between \`ComponentManagementServiceImpl\` and \`ConfigurationRowAccessors\`. Both kept local for now; refactor non-urgent.
- **P3 acknowledged:** PATCH \`versionRange\` validation was pre-existing; test added as non-regression coverage with explicit "(pre-existing)" tag in the display name.

## Test plan
- [x] \`AUTH_SERVER_URL=http://localhost:9999 AUTH_SERVER_REALM=octopus ./gradlew build -x dockerBuildImage -x dockerPushImage -x ocCreate -x ocProcess -x :components-registry-automation:test\` exits 0
- [x] \`V4WriteValidationTest\` — 8/8 green (pre-fix: 7 fail / 1 pass)
- [x] \`ListComponentsSystemFilterTest\` — 3/3 green (was 2/2 with weak assertions)
- [x] Independent Sonnet subagent review; one P1 + two P2 findings addressed before push
- [ ] CI green

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/ConfigurationRowAccessors.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/ConfigurationRowAccessors.kt
@@ -122,7 +122,7 @@ internal fun ComponentConfigurationEntity.applyScalarValue(
     clearAllScalarColumns()
 
     when (attributePath) {
-        "build.buildSystem" -> buildSystem = requireString(attributePath, value)
+        "build.buildSystem" -> buildSystem = requireBuildSystem(attributePath, value)
         "build.buildSystemVersion" -> buildSystemVersion = requireString(attributePath, value)
         "build.javaVersion" -> javaVersion = requireString(attributePath, value)
         "build.mavenVersion" -> mavenVersion = requireString(attributePath, value)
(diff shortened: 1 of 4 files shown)


pgorbachev:
Addressed in 5991ce9:

**Copilot inline:**
- L18 test: removed unused `MockMvcRequestBuilders.get` import.
- L236 service: dropped the misleading comment that claimed a top-level `validateRangeSyntax` call existed — downstream patch block already validates, no top-level guard needed.

**Sonnet 2nd-pass:**
- `PACKAGE_TYPES` (P2): added TODO comment so future maintainers know the hand-curated `{"DEB","RPM"}` set tracks the resolver's package-type emission (refactor to a shared `PackageType` enum deferred).
- `escrow.generation` (P2): same-class defect (resolver does `EscrowGenerationMode.valueOf` with silent null fallback) but not in the original 6 Copilot findings; deferring to a follow-up to keep this PR focused.
- `@DirtiesContext` (P3): kept; `@MockBean` already isolates context but the annotation is defensive-in-depth and matches the convention of sibling test classes.
- Random fixture names (P3): non-issue under normal CI (no retry), noted for future test-runner flakiness.

PR #216, opened 2026-05-16
test(import): IMP-003 regression guard — KTS-only build-block shape

pgorbachev:
## Summary

Adds a regression guard test (`IMP-003`) covering the production-DSL shape where the KTS file carries the full `build { tools { database { oracle { ... } } } }` block and the Groovy counterpart has no component-level `build` block. This matches the cluster-D production components (the ones with Oracle/PTC/PTK build-tool beans).

**This is NOT a bug fix.** PR-D+E (#208) handles this shape correctly. The test exists to lock the contract in place: if `attachBuildToolBeans` ever stops firing for this shape after a refactor, IMP-003 turns red.

## How the false alarm started

Local compat-test against the merged `feat/schema-v2-sql` showed 510 active divergences. Root cause: the local candidate JVM was in a half-broken state — two JVMs (one stale, one fresh) had been racing on the same Postgres after a partial `--reset-db`. The race produced 22 `duplicate key value violates unique constraint "components_component_key_key"` errors during auto-migrate, which made `requireMigrationSucceeded` throw → context refresh failed → partial commits left the DB inconsistent. Cards_db's BASE row got committed but the Oracle bean attach was rolled back.

After a clean reset (kill all candidate JVMs, wipe + recreate Postgres volume, start a single fresh candidate), auto-migrate completed `total=948, migrated=948, failed=0`, and `component_build_tool_beans` correctly contained 8 rows — one for each cluster-D component:

| component_key | bean_type | version_pattern |
|---|---|---|
| cards_db | oracleDatabase | `[12,)` |
| DBK | kProduct | (null) |
| DM | dDbProduct | (null) |
| dwh_db | oracleDatabase | `[12,)` |
| kernel_db | oracleDatabase | `[12,)` |
| pm_java | cProduct | (null) |
| ufxcore | cProduct | (null) |
| w4manager | kProduct | (null) |

`curl /rest/api/2/components/cards_db/versions/.../build-tools?ignore-required=true` returns `[{"@type":"oracleDatabase","type":"ORACLE","settingsProperty":"db","version":"[12,)","edition":null}]` matching baseline.

## What's in this PR

- **`Defaults.groovy`** — adds an empty `build { }` block so test components without their own component-level `build` block inherit a non-null `buildConfiguration` clone-base (matching the production Defaults shape).
- **`TestComponents.kts`** — adds `TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY` with the full KTS DSL: `productType = "PT_K"` (test-registered key), `build { tools { database { oracle { version = "12.0" } } } }`.
- **`TestComponents.groovy`** — adds the Groovy counterpart with NO component-level `build` block; two version ranges (`[2.0,)` and `(,2.0)`), per-range `build { requiredTools }`, vcsSettings, escrow.generation override — full cards_db shape.
- **`BuildToolBeansImportTest`** — IMP-003 asserts `component_build_tool_beans` has exactly 1 row (oracleDatabase 12.0) on the BASE row.

## Test plan

- [x] `./gradlew :components-registry-service-server:test --tests "*BuildToolBeansImportTest*"` — all 3 tests pass (imp001, imp002, imp003).
- [x] `--reset-db` on local candidate against current `feat/schema-v2-sql` HEAD: auto-migrate clean, all 8 cluster-D components persist their build-tool-bean rows, `/build-tools` endpoint matches baseline.

### DIFF
--- a/components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/migration/BuildToolBeansImportTest.kt
+++ b/components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/migration/BuildToolBeansImportTest.kt
@@ -137,4 +137,40 @@ class BuildToolBeansImportTest {
             "Expected PTKProductToolBean(03.49) in $buildTools"
         }
     }
+
+    // -------------------------------------------------------------------------
+    // IMP-003: cards_db-shape — KTS carries build.tools, Groovy has NO build block
+    // -------------------------------------------------------------------------
+
+    @Test
+    @DisplayName(
+        "IMP-003: TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY (KTS-only build block, Groovy without build) " +
+            "MUST persist its Oracle bean after auto-migrate — reproduces production cards_db shape",
+    )
+    @Transactional
+    fun imp003_ktsOnlyBuildBlock_persistsOracleBean() {
+        val component = componentRepository.findByComponentKey("TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY")
+        assertNotNull(component, "TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY must be migrated")
+
+        val baseRow = configurationRepository.findBaseByComponentId(component!!.id!!)
+        assertNotNull(baseRow, "BASE row must exist for TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY")
+
+        val beans = buildToolBeanRepository
+            .findByComponentConfigurationId(baseRow!!.id!!)
+            .sortedBy { it.sortOrder }
+
+        assertEquals(
+            1,
+            beans.size,
+            "Expected 1 build-tool bean row (oracleDatabase 12.0); got ${beans.size}. " +
+                "The KTS-only-build-block shape (matching production cards_db) MUST flow " +
+                "through the merge so that ImportServiceImpl.attachBuildToolBeans sees " +
+                "the Oracle bean. If this test is red, the merge is short-circuiting " +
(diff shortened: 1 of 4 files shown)


pgorbachev:
Closing — the originally suspected bug (cards_db Oracle bean missing in DB after local compat run) turned out to be **environment chaos**, not a real PR-D+E regression.

Root cause of the false alarm: two candidate JVMs (stale PID 17799 + fresh PID 75123) were racing on the same Postgres after a partial `--reset-db`, producing 22 duplicate-key clashes on `components.component_key` → `requireMigrationSucceeded` threw → context refresh failed → partial commits left the DB inconsistent (cards_db BASE row committed, Oracle bean attach rolled back).

After a clean reset (kill all JVMs, wipe + recreate Postgres volume, start a single fresh candidate), auto-migrate completed `total=948, migrated=948, failed=0`. All 8 cluster-D components have their proper build-tool-bean rows (cards_db / dwh_db / kernel_db → oracleDatabase, DBK / w4manager → kProduct, DM → dDbProduct, pm_java / ufxcore → cProduct). `curl /build-tools` on cards_db returns the expected Oracle bean matching baseline.

Trade-off: this PR's regression-guard test fixture (new `TEST_COMPONENT_BUILD_TOOLS_KTS_ONLY` in global `TestComponents.{kts,groovy}` + empty `build {}` in Defaults) breaks 3 client tests that hardcode component counts (`testGetAllComponents`, RES-001 jira-component-version-ranges, RES-003 product-mapping). Updating those counts would touch ~10 assertions across unrelated test surfaces just to add a regression guard for behavior that already works.

If we want regression coverage later, the right shape is a **focused unit test** that constructs the relevant config objects in-memory and calls `attachBuildToolBeans` directly — no global fixture changes.

PR #222, opened 2026-05-16
fix(compat): kill Set-positional false-positives at raw layer (cluster I)

pgorbachev:
## Summary
- Today's compat run reports ~1781 STRUCTURAL_DIFFs on Set-shape endpoints. The diffs are infra false-positives: `JsonShape.diff` walks arrays positionally while `Set` element wire-order is non-deterministic across stands.
- Introduce `RawArraySorters` with a narrow per-endpoint registry of stable sort-key extractors, and pre-sort both sides in `CompatibilityTestBase.compareRaw` before structural diff.
- Pass-through identity for unregistered endpoints — opt-in surface, no implicit auto-detection.

## Registered endpoints (deliberately narrow)

| Endpoint | Sort key | Notes |
|---|---|---|
| `GET /rest/api/2/common/jira-component-version-ranges` | `componentName` + NUL + `versionRange` | `componentName` is the TOP-LEVEL field on `JiraComponentVersionRangeDTO`; the nested `component` is a `JiraComponentDTO` with no `id`. |
| `GET /rest/api/2/projects/{projectKey}/jira-component-version-ranges` | (same) | Same DTO, filtered by `projectKey` server-side. Shares the key extractor with the global endpoint. |
| `GET /rest/api/3/components` | `component.id` | `ComponentV3.component` is `ComponentV2` with `id` field. |

## Behaviour
| Case | Before | After |
|---|---|---|
| Registered Set endpoint, same elements / different wire-order | ARRAY_SIZE / TYPE_MISMATCH / KEY_MISSING flood | 0 shape diffs |
| Registered endpoint, true size mismatch | ARRAY_SIZE_MISMATCH | ARRAY_SIZE_MISMATCH (unchanged) |
| Registered endpoint, per-element shape diff | KEY_MISSING_* | KEY_MISSING_* (unchanged) |
| Unregistered endpoint | shape diff if any | identity pass-through (unchanged) |
| Non-array root | unchanged | unchanged (identity) |

## Test plan
- [x] New `RawArraySortersTest` with 8 unit tests:
  - **Heterogeneous-shape regression tests** (jira global + jira per-project + v3): one element carries an optional field (`displayName` / `name`), the other does not. Positional diff would surface `KEY_MISSING_*`; zero-diff is reached ONLY after key-correct alignment. These would FAIL under a no-op sorter or under the previous buggy `component.id` key (which collapsed to `""`).
  - Identity pass-through (unregistered endpoint, non-array root) using `assertSame`.
  - Anti-regression: true size mismatch and per-element shape divergence still surface.
- [x] Full `./gradlew build` (with standard exclusions) passes.
- [ ] Next compat run shows zero raw `STRUCTURAL_DIFF` records on the three registered endpoints.

## Review-feedback log
- `1b8b5f0` `test:` introduce helper + tests.
- `ac143af` `fix:` wire into `compareRaw`.
- `08b6e1d` `docs:` Sonnet review fixes — NUL separator via `KEY_SEP`, read-only contract, `arrayNode(int)` clarification.
- `36f4ec0` `fix:` Copilot P0 — correct `componentName` (top-level) instead of non-existent `component.id`; rewrite test fixture to mirror wire shape; add regression-guard test.
- `ab86dc3` `fix:` external reviewer P2 — register `projects/{projectKey}/jira-component-version-ranges`; rewrite tests with heterogeneous-shape elements so a no-op sorter would fail; soften the `Set`-iteration comment to match the actual contract.

Plan reference: `~/.claude/plans/image-6-playful-gizmo.md` TASK-3.

### DIFF
--- a/components-registry-compat-test/src/test/kotlin/org/octopusden/octopus/components/registry/compat/CompatibilityTestBase.kt
+++ b/components-registry-compat-test/src/test/kotlin/org/octopusden/octopus/components/registry/compat/CompatibilityTestBase.kt
@@ -162,7 +162,14 @@ abstract class CompatibilityTestBase {
 
         // Skip shape diffing if status codes already diverged or no JSON
         if (baseline.status == candidate.status && baseline.json != null && candidate.json != null) {
-            val shapeDiffs = JsonShape.diff(baseline.json, candidate.json)
+            // For Set-shape endpoints (`/jira-component-version-ranges`, `/v3/components`)
+            // the wire-order of elements is non-deterministic across stands. Pre-sort
+            // both sides by a stable per-endpoint key so JsonShape.diff doesn't report
+            // positional false-positives. Pass-through for unregistered endpoints
+            // (see RawArraySorters and its unit test for the registered list + contract).
+            val baselineForShape = RawArraySorters.stableSorted(endpoint, baseline.json)
+            val candidateForShape = RawArraySorters.stableSorted(endpoint, candidate.json)
+            val shapeDiffs = JsonShape.diff(baselineForShape, candidateForShape)
             for (sd in shapeDiffs) {
                 categories += DiffClassifier.STRUCTURAL_DIFF
                 DiffCollector.record(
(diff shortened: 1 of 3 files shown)


pgorbachev:
Addressed both Copilot review comments in `36f4ec0`:

- **Sort key**: `component.id` does not exist on `JiraComponentVersionRangeDTO` — the identifier is `componentName` at the top level. Switched the key to `componentName + NUL + versionRange`.
- **Test fixture**: rewrote `rangeEntry` to mirror the wire shape (`componentName` top-level, nested `component` as `JiraComponentDTO` with `displayName`).
- Added `jiraComponentVersionRanges_distinctComponentNameSameRange_zeroShapeDiffs` regression test that would fail under the previous buggy key.

Full `./gradlew build` passes.

PR #225, opened 2026-05-16
fix(schema-v2): preserve V1 wildcard artifactPattern when DSL omits artifactId (RES-C-prime)

pgorbachev:
## Summary
PR #209 (per-range maven-artifacts override leak fix) inadvertently changed the public-API semantic of `/maven-artifacts` for any DB-routed component whose DSL omits an explicit `artifactId` line: it unconditionally synthesized artifactIds from per-config `distribution.GAV()`. V1 in-memory resolver returns `EscrowModuleConfig.artifactIdPattern` verbatim — for components without an explicit declaration, that's the wildcard literal `[\w-\.]+` inherited from `Defaults.artifactId = ANY_ARTIFACT`. 109 components in today's compat run show the divergence.

Implements **Contract B** from `~/.claude/plans/image-6-playful-gizmo.md` TASK-1:
- Gate the GAV-extraction branch on the presence of a per-range `DISTRIBUTION_MAVEN` MARKER row for that range.
- Non-marker ranges return `componentLevelFallback` from `componentEntity.artifactIds` — the row(s) `ImportServiceImpl.kt:893-902` populates from `EscrowModuleConfig.artifactIdPattern`.

## Behaviour
| Range shape | Before (PR-C) | After (PR-C-prime) |
|---|---|---|
| Range has `DISTRIBUTION_MAVEN` MARKER row | GAV-extracted `(group, CSV)` from marker | GAV-extracted `(group, CSV)` from marker — unchanged, RES-C bug stays fixed |
| Range has no marker, component has explicit `artifactId` in DSL | GAV-extracted CSV (wrong if DSL artifactId differs from GAV) | Component-level fallback (`(group, artifactId)`) — V1 parity |
| Range has no marker, no explicit `artifactId` | GAV-extracted CSV (wrong: 109 components in prod) | Component-level fallback (`(group, "[\w-\.]+")`) — V1 parity |
| No component-level fallback AND no marker | Component-level fallback null → filtered out | (unchanged) |

## Test plan
- [x] **RES-C-001/002/003** stay GREEN (fixture updated to include `componentEntity.artifactIds` to mirror production import shape).
- [x] **RES-C-004** new: single range, BASE has per-config GAV, no marker → V1 wildcard wins.
- [x] **RES-C-004b** new: two ranges via `RANGE_PRESENCE`, no markers → both ranges return wildcard.
- [x] **RES-C-005** new: mixed — BASE+GAV, MARKER on range B only → A=fallback, B=GAV.
- [x] Full `./gradlew build` passes.

## Review-feedback log
- `c76a863` `test:` introduce 3 RED tests + helpers (`addComponentLevelArtifact`, `makeRangePresenceRow`).
- `78adb80` `fix:` Contract B implementation + RES-C-001/002/003 fixture updates to production shape.
- `c840e89` `docs:` Sonnet review fixes — `ROW_TYPE_MARKER` constant, sentinel-comment on `filterValues`, RES-C-006 TODO note.

Plan ref: `~/.claude/plans/image-6-playful-gizmo.md` TASK-1.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/DatabaseComponentRegistryResolver.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/DatabaseComponentRegistryResolver.kt
@@ -271,11 +271,18 @@ class DatabaseComponentRegistryResolver(
                 ?: throw NotFoundException("Component '$component' is not found")
 
         // Component-level fallback: the top-level groupId/artifactId rows imported from DSL.
-        // For components whose DSL omits an explicit `artifactId` line, ImportServiceImpl
-        // writes the inherited `Defaults.artifactId = ANY_ARTIFACT (/[\w-\.]+/)` verbatim
-        // (see ImportServiceImpl.kt:893-902). The V1 in-memory resolver returns that
-        // wildcard literal as-is; we must do the same to preserve API parity (RES-C-prime).
-        val artifactIdRows = componentEntity.artifactIds.toList()
+        // For components whose DSL omits an explicit `artifactId` line, the import path
+        // writes the inherited `Defaults.artifactId = ANY_ARTIFACT (/[\w-\.]+/)` verbatim;
+        // the V1 in-memory resolver returns that wildcard literal as-is (RES-C-prime).
+        //
+        // `componentEntity.artifactIds` is a JPA `@OneToMany` without `@OrderBy`, so the
+        // DB load order is not contractual. Sort by `id` (UUID) for a deterministic
+        // CSV across reloads of the same DB state. Note: the import writes one row per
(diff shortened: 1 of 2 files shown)


pgorbachev:
Addressed all 3 Copilot inline comments in d6af9b1:

- **`@OrderBy` / non-deterministic CSV** (P0): `componentEntity.artifactIds` now sorted by `id` (UUID) in the resolver — stable across reloads. `first().groupPattern` is safe because import always writes rows with a SHARED `groupPattern` (one DSL `groupId` per component). DSL-order preservation for multi-artifact CSVs is filed as a follow-up.
- **Line-number drift in `ROW_TYPE_MARKER` KDoc** (P2): removed the brittle line references.
- **Local plan path in test comment** (P2): replaced `~/.claude/plans/...` with the in-tree `RES-C-prime` identifier.

Tests 6/6 GREEN.

PR #227, opened 2026-05-16
fix(schema-v2): preserve DSL artifact-id order for multi-artifact CSV (TD-008 / VAL-006)

pgorbachev:
## Summary
TC build 3456 (`fix/artifact-ids-sort-order` predecessor or 2.0.84-3456) surfaced a real failure of `GitVsDbValidationTest.VAL-006`: V1 returns the raw DSL `artifactIdPattern` string verbatim (e.g. \`art-core,art-logging,art-http,art-xml,art-cli\`), V2 reloaded the multi-row \`componentEntity.artifactIds\` collection in UUID order and rejoined the CSV out of order.

PR-C-prime (#225) introduced a sort-by-UUID step for determinism *across reloads*, but UUIDs are random, so the joined CSV did not match V1's DSL declaration order.

## Fix
- Schema: add \`sort_order INT NOT NULL DEFAULT 0\` to \`component_artifact_ids\` (V1__schema.sql edited in place per the migration policy — no incremental migration; re-import is the data-migration SoT).
- Entity: add \`sortOrder: Int = 0\` to \`ComponentArtifactIdEntity\`.
- Import: \`ImportServiceImpl\` writes one row per CSV token with \`sortOrder = index\` in DSL declaration order. \`ComponentManagementServiceImpl.addArtifactIds\` (V4 admin write path) uses \`forEachIndexed\` for the same reason.
- Read: \`DatabaseComponentRegistryResolver\` sorts rows by \`sortOrder\` ASC before joining the CSV — matches V1 byte-for-byte.

## Test plan
- [x] New **RES-C-006**: shuffled insertion order asserts \`sort_order\`-preserving join (would fail under previous UUID-sort).
- [x] RES-C-001/002/003/004/005 stay GREEN.
- [x] Full \`./gradlew build\` passes locally.
- [ ] Expect TC \`GitVsDbValidationTest.VAL-006\` GREEN on this branch.

Plan ref: \`~/.claude/plans/image-6-playful-gizmo.md\` TD-008.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/entity/ComponentArtifactIdEntity.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/entity/ComponentArtifactIdEntity.kt
@@ -26,4 +26,7 @@ class ComponentArtifactIdEntity(
 
     @Column(name = "artifact_pattern", columnDefinition = "TEXT", nullable = false)
     var artifactPattern: String = "",
+
+    @Column(name = "sort_order", nullable = false)
+    var sortOrder: Int = 0,
 )
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -588,12 +588,13 @@ class ComponentManagementServiceImpl(
         component: ComponentEntity,
         requests: List<org.octopusden.octopus.components.registry.server.dto.v4.ArtifactIdRequest>,
     ) {
-        requests.forEach { req ->
+        requests.forEachIndexed { index, req ->
             component.artifactIds.add(
                 ComponentArtifactIdEntity(
                     component = component,
                     groupPattern = req.groupPattern,
                     artifactPattern = req.artifactPattern,
+                    sortOrder = index,
                 ),
             )
         }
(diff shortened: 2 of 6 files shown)


pgorbachev:
Addressing both Copilot findings:

**P0 — Flyway forward-only / checksum validation** (V1__schema.sql:280)

This is the schema-v2 in-place-edit override of the generic forward-only policy, NOT a violation. Context:

- Schema-v2 is NOT yet in production (`feat/schema-v2-sql` is the integration branch).
- The candidate config uses `spring.flyway.baseline-on-migrate: true` and the local-stand / QA workflow wipes the Postgres volume before redeploy (see `scripts/local-stands/verify.sh --reset-db`).
- Per the migration policy, V1__schema.sql is edited in place until first prod cut-over, and the Groovy DSL→DB import is the data-migration source-of-truth. Incremental V2/V3 migrations only start once V1 is frozen at prod-deploy time.
- The doc at `docs/db-migration/non-functional-spec.md:137-139` describes the steady-state contract that applies once the schema is frozen.

If we adopted incremental migrations here, every schema fix on the integration branch (already several this week — `build.buildTools` table, `vcs_settings_entries.name` NOT NULL tightening, etc.) would have ballooned into a V2+ file each. That's the trade-off the policy made.

**P2 — sortOrder ties non-deterministic** (DatabaseComponentRegistryResolver.kt:284)

Valid concern for a future write path that forgets to populate sortOrder. Addressed in the follow-up commit — switched from `.sortedBy { it.sortOrder }` to:

```kotlin
.sortedWith(compareBy({ it.sortOrder }, { it.artifactPattern }))
```

Secondary key on `artifactPattern` (alphabetical) gives a deterministic tiebreak. Today the import + V4 admin paths both set sortOrder explicitly, so ties don't occur in practice — this is defensive-depth for future code paths.

PR #248, opened 2026-05-17
chore: post-#247 review followup — anchor wording + error-msg truncation

pgorbachev:
Small post-merge cleanups from the Opus adversarial review of PR #247:

- **P1-B** `ImportServiceImpl.kt:1486-1500` — anchor comment rewritten. UUID-strategy wasn't the actual safety property; reference identity from `SimpleJpaRepository.save` is. Updated to describe the real invariant + real regression class.
- **P2-A** `ComponentSourceRegistryImpl.kt` — truncate echoed `source` in the `IllegalArgumentException` message to 80 chars with `…` marker. Defensive against a future V4 endpoint exposing user-controlled `source` (none today; TD-011 might add one).
- Style: dropped redundant double-`private` on the companion; added `MAX_ERROR_ECHO_CHARS` constant.

New test `setComponentSource_oversizedRejected_messageTruncated` (8 total in the class) pins the truncation.

## Test plan

- [x] `./gradlew :components-registry-service-server:test --tests *ComponentSourceRegistryWhitelistTest*` — 8/8 pass.
- [ ] CI green on push.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentSourceRegistryImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentSourceRegistryImpl.kt
@@ -32,7 +32,12 @@ class ComponentSourceRegistryImpl(
         // silently exclude it from getDbComponentNames() → 404 blackhole.
         val normalized = source.trim().lowercase()
         require(normalized in ALLOWED_SOURCES) {
-            "Invalid component source '$source'; allowed values: ${ALLOWED_SOURCES.joinToString(", ")}"
+            // Truncate echoed value to avoid log-injection / oversized payload landing
+            // verbatim in ControllerExceptionHandler's HTTP response body and the WARN
+            // log line (PR #247 Opus review P2-A). Hard ceiling of MAX_ERROR_ECHO_CHARS.
+            val echo = source.take(MAX_ERROR_ECHO_CHARS)
+            val ellipsis = if (source.length > MAX_ERROR_ECHO_CHARS) "…" else ""
+            "Invalid component source '$echo$ellipsis'; allowed values: ${ALLOWED_SOURCES.joinToString(", ")}"
         }
         val entity =
             componentSourceRepository.findById(name).orElse(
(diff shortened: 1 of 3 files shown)


components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentSourceRegistryImpl.kt
```diff
@@ -32,7 +32,12 @@ class ComponentSourceRegistryImpl(
         // silently exclude it from getDbComponentNames() → 404 blackhole.
         val normalized = source.trim().lowercase()
         require(normalized in ALLOWED_SOURCES) {
-            "Invalid component source '$source'; allowed values: ${ALLOWED_SOURCES.joinToString(", ")}"
+            // Truncate echoed value to avoid log-injection / oversized payload landing
+            // verbatim in ControllerExceptionHandler's HTTP response body and the WARN
+            // log line (PR #247 Opus review P2-A). Hard ceiling of MAX_ERROR_ECHO_CHARS.
+            val echo = source.take(MAX_ERROR_ECHO_CHARS)
```
pgorbachev: [P2] This only limits length, but the comment says the goal is also to avoid log-injection. A short rejected value like `bad\nWARN forged` is under 80 chars, so it still lands verbatim in `ControllerExceptionHandler`'s WARN line and response body with the newline intact. Please normalize CR/LF/control chars before echoing, e.g. replace them with spaces or escaped sequences, and add a test for a short newline-injected source.

pgorbachev (commented):
Reviewed locally. I left one inline comment; checks are green.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentSourceRegistryImpl.kt
pgorbachev: Fixed in `a02eb43b`. Added `sanitizeForEcho()` in the companion that replaces every ISO C0/C1 control char (0x00..0x1F and 0x7F..0x9F) with `\xNN` BEFORE truncation, so `bad\nWARN forged` no longer lands verbatim. Two new pinned tests:
- `setComponentSource_newlineInjected_sanitized` — short `bad\nWARN forged` → no raw `\n`, `\x0A` present.
- `setComponentSource_assortedControls_sanitized` — CR/TAB/DEL/NEL all escaped.

Truncation now operates on the sanitized output, so no partial escape can be torn at the cap boundary.

PR #302, opened 2026-05-27
fix(v4): preserve per-range hotfixVersionFormat through schema-v2 import + resolver

pgorbachev:
## Summary

- v3's schema-v2 import was discarding per-range `jira.componentVersionFormat.hotfixVersionFormat` DSL overrides — only the per-component scalar was written. The DB-backed resolver then surfaced the same hotfixVersionFormat for every version of a component, regardless of any per-range `componentVersion(R) { jira { hotfixVersionFormat = … } }` blocks. v2.0.87 prod (= `f1-gateway` baseline) propagates the per-range value correctly through the upstream Groovy library; v3 candidate was diverging.
- Fix is additive (mirrors how `releaseVersionFormat`/`buildVersionFormat`/etc. are stored): adds a per-range `component_configurations.jira_hotfix_version_format` column; keeps the per-component `components.jira_hotfix_version_format` as the inheritance fallback (Defaults / top-level DSL). Resolver reads per-range first, falls back to per-component, then to empty string.
- V4 editor CRUD per-range write extension is intentionally out of scope — DSL import is the only producer in production today. `JiraAspectResponse.hotfixVersionFormat` is exposed read-only; `JiraAspectRequest` is unchanged.

## Commit shape (strict TDD)

1. `test(v4): pin per-range hotfixVersionFormat override regression (RED)` — new `TEST_PER_RANGE_HOTFIX_FORMAT` fixture + failing UT against the DB-backed profile; captured failure JSON in the commit body shows the bug. Boundary test not in this commit (per the TDD-discipline note about boundary-pin separation).
2. `fix(v4): preserve per-range hotfixVersionFormat through schema-v2 import + resolver` — schema column + entity + accessors + mapper + import + V4 read-only DTO + supporting test-data updates (MetricsTest count 56→57; expected-data JSON adds 2 entries for the new fixture; MigrationIntegrationTest RANGE_PRESENCE invariant extended to cover the new column).
3. `test(v4): boundary pin — out-of-range version falls back to base hotfixVersionFormat` — companion UT for the inheritance-fallback path, confirms the per-component base is preserved for versions outside any override range.

## Schema migration

`V1__schema.sql` is edited in-place (no new V<n> file) per the pre-prod fresh-DB-on-deploy convention. One new column added to `component_configurations`; the existing all-typed-cols-NULL CHECK on MARKER/RANGE_PRESENCE rows is extended to cover it. The DB CHECK constraint allowlist for SCALAR_OVERRIDE uses `NOT IN (<markers>)`, so no allowlist change is needed.

## Test plan

- [x] `:components-registry-service-server:test` (single-worker run; full suite — parallel runs hit Spring `DefaultCacheAwareContextLoaderDelegate` init errors due to resource exhaustion on the local stand, not code-related).
- [x] `:components-registry-service-light-client:test`.
- [x] Targeted tests confirmed GREEN: `testPerRangeHotfixVersionFormatOverridesBase_dbMode`, `testVersionOutsidePerRangeHotfixOverrideFallsBackToBase_dbMode`, `MigrationIntegrationTest`.
- [x] Sonnet code-review subagent (high effort) on the 3-commit diff — flagged one critical (`MigrationIntegrationTest` RANGE_PRESENCE invariant missing the new column) which is now addressed in commit 2, plus 2 documentation polish items in `JiraAspectResponse` KDoc and the fixture comment. No call-site for the new column was missed compared to the `jiraReleaseVersionFormat` sibling pattern; the deliberate V4 CRUD scope-omission has explanatory inline comments at the `applyBaseConfigurationCreate` and `applyBaseConfigurationPatch` call sites.
- [x] Forbidden-token grep against diff + commit messages — clean.
- [ ] CI compat-test job (post-merge): the 11 active diff records on build 2.0.84-3681 should resolve after a fresh stand rebuild against this branch.

## Why two columns, not one

The schema design distinguishes "Defaults / top-level inherited value" (per-component column on `components`) from "DSL-declared per-range override" (per-range column on `component_configurations`). The resolver layering preserves the current V4 UI semantics for the per-component hotfix-format edit while letting DSL imports carry per-range overrides through to the wire DTO.

Closes the v3 compat-job regression captured by build 2.0.84-3681 (manual curl against `f1-gateway` prod and `f1-gateway-test` candidate stands confirmed the divergence).

### DIFF
--- a/components-registry-service-client/src/test/kotlin/org/octopusden/octopus/components/registry/client/ComponentRegistryServiceClientTest.kt
+++ b/components-registry-service-client/src/test/kotlin/org/octopusden/octopus/components/registry/client/ComponentRegistryServiceClientTest.kt
@@ -155,9 +155,10 @@ class ComponentRegistryServiceClientTest : BaseComponentsRegistryServiceTest() {
 
     @Test
     fun testGetAllComponents() {
-        // 56 pre-MIG-041 + 2 new aggregator-handling fixtures (TEST_AGGREGATOR_FAKE +
-        // TEST_AGGREGATOR_MEMBER) which inherit the Defaults `system = "NONE"`.
-        assertEquals(58, componentsRegistryClient.getAllComponents().components.size)
+        // 56 pre-MIG-041 + 2 aggregator-handling fixtures (TEST_AGGREGATOR_FAKE +
+        // TEST_AGGREGATOR_MEMBER) + 1 task-#16 fixture (TEST_PER_RANGE_HOTFIX_FORMAT)
+        // which all inherit the Defaults `system = "NONE"`.
+        assertEquals(59, componentsRegistryClient.getAllComponents().components.size)
         assertEquals(
             3,
             componentsRegistryClient.getAllComponents("ssh://hg@mercurial/technical", null).components.size,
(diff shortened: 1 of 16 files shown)


pgorbachev (commented):
Code review, code-only. One issue found:

[P2] `jira.hotfixVersionFormat` null-clear scalar overrides are not honored by the resolver.

`jira.hotfixVersionFormat` is now included in `SCALAR_ATTRIBUTE_PATHS`, so it participates in the existing import contract documented in `ConfigurationRowAccessors.kt`: import may emit a `SCALAR_OVERRIDE` row with `overridden_attribute = "jira.hotfixVersionFormat"` and `jira_hotfix_version_format = NULL` to explicitly clear an inherited/base scalar for a version range. `ComponentConfigurationView.applyScalarOverride` does apply that row by setting `merged.jiraHotfixVersionFormat = null`, but `buildJiraComponent` then does `merged.jiraHotfixVersionFormat ?: component.jiraHotfixVersionFormat ?: ""`. That fallback makes the cleared range inherit the component-level base value again, so the explicit null-clear is lost.

This differs from the generic scalar override semantics already pinned for other fields. A version range that clears the hotfix format should resolve to the no-format fallback, not back to `components.jira_hotfix_version_format`. If null-clear is not meant to be supported for this field, it should not go through the generic `SCALAR_ATTRIBUTE_PATHS` / `collectScalarDiffs` null-clear path. Otherwise the resolver needs to track presence of a `jira.hotfixVersionFormat` override separately from the nullable value, and this PR should add a regression test for the null-clear case.

Relevant lines: `components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/ConfigurationRowAccessors.kt:22`, `.../ConfigurationRowAccessors.kt:62`, `components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/EntityMappers.kt:507`, `.../EntityMappers.kt:763`.

I did not run tests locally; reviewed the patch by code.

PR #308, opened 2026-06-01
feat: phase-aware updateCache + [1.8] git-mode no-migration compat gate

pgorbachev:
## What

Hardens the v3 cutover with two linked changes:

1. **Phase-aware `PUT /service/updateCache`.** While git-sourced components remain (`migration-status git > 0`) the endpoint re-reads the Git config and returns **200 + the refresh duration (ms)** — byte-identical to the pre-v3 (2.0.87) endpoint — so v1/v2/v3 keep behaving like the old version throughout the hybrid period. It returns **410 Gone** only once everything is migrated (`git <= 0`), and **409** while a `COMPONENTS` migration is running. The 410 is the cutover *end-state*, not an unconditional retirement.

2. **`[1.8]` no-migration (git-mode) compatibility build (`id18`).** Runs the v1/v2/v3 compat suite against a candidate booted with no migration (`default-source=git`), asserting the **deploy-without-migration no-op invariant**: deploying v3 and not migrating must be byte-identical to 2.0.87 for v1/v2/v3. Its `known-deltas-git.json` is intentionally empty — any active diff there is a real regression.

## How

- `ComponentsRegistryServiceController.updateConfigCache()` is phase-aware, gated on `ImportService.getMigrationStatus().git` (`= gitResolver.size − countBySource("db")`) and `MigrationLifecycleGate`. New unit test covers all four branches; the existing integration test was flipped (410 → 200).
- Compat known-deltas split into `known-deltas-db.json` (migrated candidate; the `updateCache` 410 delta) + empty `known-deltas-git.json`, selected by `-Pcompat.known-deltas` (default db).
- `scripts/local-stands/teamcity-run.sh` gains `CANDIDATE_MODE` (`db`|`git`); git-mode uses highest-precedence CLI overrides (`--components-registry.auto-migrate=false --components-registry.default-source=git`, which beat profile YAML and service-config), a mode-aware `defaultSource` precondition, and always passes `-Pcompat.candidate.mode`.
- `.teamcity/settings.kts`: `id18` cloned from `id17` (only delta: `CANDIDATE_MODE=git`); both serialized via a `crs-local-stand` shared-resource write-lock (they share fixed ports + the local Postgres compose).
- The heavy/mutating `updateCache` case is excluded from the git-mode read-parity surface (it can transport-fail under parallel load; its git-mode 200 behaviour is already unit/integration-tested).
- Docs: `docs/registry/api-compat-deltas.md`, ADR-013.

## Verification

- Full `gradlew build` green (run twice, pre- and post-fixup; ~17 min each).
- Independent review: correctness pass + adversarial pass → fixups → focused re-review, all green.

## ⚠️ Reviewer — two items that cannot be validated locally

1. **TeamCity DSL** (`sharedResource` / `sharedResources`) compiles only against the TC server's DSL plugin repo — please confirm on "Load settings from VCS". Fallback if the import path is wrong on this server: drop the explicit `projectFeatures.sharedResource` import (the `kotlin.*` wildcard already covers it) and/or re-check `buildFeatures.sharedResources`.
2. **`id18` git-mode RED-then-GREEN demo** needs the 2.0.87 baseline + candidate JARs — run `id18` on TeamCity (or locally via `CANDIDATE_MODE=git bash scripts/local-stands/teamcity-run.sh` with the JARs) to confirm `0 active diffs` and that `known-deltas-git.json` stays empty.

## Merge order

Rebase on `v3` **after** the separate ~119-db-mode-diffs fix branch lands. Those diffs are db-mode (DB-resolver vs baseline) and are **not** introduced here — `[1.7]`/`id17` is expected to stay red on this branch until that work lands; acceptance here is gated on `[1.8]`/`id18` (git-mode) + the server unit tests.

## Release Notes

* **New Features**
  * The `PUT /rest/api/2/components-registry/service/updateCache` endpoint now implements phase-aware response behavior, returning HTTP 200 with refresh duration when Git components are present, and HTTP 410 once fully migrated.
  * Added git-mode compatibility testing capability to validation suite.

* **Documentation**
  * Updated API contract and migration strategy documentation to reflect phase-aware endpoint behavior and per-mode compatibility expectations.


pgorbachev:
@coderabbitai review

pgorbachev:
Thanks @coderabbitai — all three are trivial nitpicks; dispositions:

1. **id17/id18 duplication (.teamcity/settings.kts).** Deliberate: each build object in this file is standalone (id15/id16/id17 already duplicate), and cloning avoids destabilizing the proven id17 — id18 differs only by id/name, the work-dir namespace, and the single `CANDIDATE_MODE=git` export. Extracting shared step-builders is a reasonable future refactor but out of scope here.

2. **build.gradle git-mode/known-deltas pairing.** Confirmed correct for CI: `teamcity-run.sh` always pairs `-Pcompat.candidate.mode=git` with `-Pcompat.known-deltas=known-deltas-git.json`, guarded by the `test-teamcity-run.sh` self-test. The fail-fast guard only helps ad-hoc *manual* gradle runs; noting as a low-priority follow-up rather than blocking.

3. **Concurrent-refresh 409 coverage.** That branch is `AtomicBoolean.compareAndSet` (JDK-tested); deterministic unit coverage needs a visibility relaxation or a concurrency harness. Accepted as a nitpick — the migration-409 path is covered.

The two actionable (Copilot) findings — concurrent-refresh serialization and the error-masking false-410 — are fixed in 87ae7746. Full `gradlew build` green.

PR #311, opened 2026-06-01
R2: distribution + multi-value extended filters & in-use meta endpoints (SYS-045/046)

pgorbachev:
## What

CRS-side of extended-search **Round 2**, on top of `v3`. Two parts.

### A — distribution filters (SYS-045)
`?distributionExplicit=` / `?distributionExternal=` were wired into the DTO + controller in an earlier WIP commit but had **no `buildSpecification` predicate**, so they were silently inert (returned the full list). This adds the two scalar-equal predicates, mirroring `solution` (nullable columns → `=false` matches only explicit-false rows; `NULL` rows are excluded). Test-first (red → green).

### B — multi-value extended filters + in-use meta endpoints (SYS-046)
`clientCode`, `jiraProjectKey`, `parentComponentName`, `groupKey` become **multi-value exact `IN`** filters (CSV / repeatable params, normalised through the existing `normalizeCsvParam`), parity with `owner` / `system` / `buildSystem` / `labels`.

> ⚠️ **Intentional behaviour change:** `clientCode`, `jiraProjectKey`, `groupKey` move from substring `LIKE` to **exact match** — the Portal is switching these to multi-select dropdowns sourced from real in-use values, where exact OR is the right semantics. `parentComponentName` widens from single exact to multi exact. `ComponentFilter` is server-internal, not an external wire type; single-value still works as a degenerate one-element `IN`.

Four companion **in-use** option-list endpoints (gated by `ACCESS_COMPONENTS`, parity with `/meta/owners`), each sorted / distinct / null-blank-filtered:
- `/meta/client-codes`, `/meta/jira-project-keys` (BASE row)
- `/meta/parent-component-names` — only keys actually referenced as a parent (NOT the can-be-parent candidate set the editor picker uses via `?canBeParent=true`)
- `/meta/group-keys` — only groups with ≥1 member (no dead options)

New repo queries: `ComponentRepository.findDistinctClientCodes` / `findDistinctJiraProjectKeys` / `findDistinctParentComponentNames`; `ComponentGroupRepository.findDistinctGroupKeys`.

## Tests
- `ListComponentsExtendedFiltersTest` — distribution `true` / `false` / `NULL`-exclusion cases; the four filters updated to exact-IN + multi-value + negative assertions.
- `MetaInUseOptionsEndpointsTest` (new) — the four endpoints: distinct / sorted / in-use-only / no-dead-options + always-200-array shape.
- `TeamcitySyncServiceTest` stub implements the new repo methods.

## Verification
- Full `./gradlew build` (docker + automation excluded; compat-test self-skips) — green.
- `qualityStatic` — green.
- Independent subagent review — no blockers; the three follow-up test-coverage gaps it raised are closed in the final commit.

## Docs
`docs/registry/requirements-common.md` (SYS-045 / SYS-046) + `docs/registry/functional-spec.md` §1.1.

## Not in this PR
The **Portal** side of R2 (distribution tri-state toggles + multi-select meta dropdowns) is a deliberate follow-up: it's a coordinated CRS↔Portal contract break that must not deploy before this ships and `crs.version` is bumped.

## Release Notes

* **New Features**
  * Added multi-value filtering for component attributes: client codes, JIRA project keys, parent component names, and group keys—now supports filtering by multiple values simultaneously.
  * Added new distribution-type filters: explicit and external distribution options.
  * Added metadata endpoints for discovering available filter values across client codes, JIRA project keys, parent component names, and group keys.

* **Improvements**
  * Component filters now use exact-match semantics instead of substring matching for precise results.


pgorbachev:
@coderabbitai review

pgorbachev:
@coderabbitai review

PR #315, opened 2026-06-02
feat(compat): replay real captured POST bodies in [1.7]/[1.8]

pgorbachev:
## What

Replays the **real captured prod POST bodies** (`crs-compat-trace/post-bodies.ndjson`) against the baseline + candidate stands, wiring it into the `[1.7]` (id17, db-mode) and `[1.8]` (id18, git-mode) local-stand builds and the manual `[1.6]` (id16).

Until now `TraceReplayCompatTest` synthesised POST bodies (`BodyFixtures`), so the body-driven endpoints (`find-by-artifacts`, `detailed-versions`) only exercised a trivial 1-element lookup. The real bodies are the actual multi-artifact / multi-version batches clients send — that is where schema-v2 resolution divergences actually surface.

## Changes

- **`RealBodyReplayCompatTest`** (HTTP-tagged via `CompatibilityTestBase`): reads `compat.bodies.file` / `COMPAT_BODIES_FILE`, parses `{count,method,path,body}` NDJSON, POSTs each real body to both stands, diffs via the shared `Comparators.compareRaw` pipeline. Reuses `TraceReplayCompatTest.parsePathAndQuery`. Diffs flow into the same `DiffCollector` → `compatibilityReporter` gate as every other compat test.
- **`RealBodyReplayParseTest`** (`@Tag("unit")`): 12 parser cases, runs in `:unitTest` on every PR.
- **`build.gradle`**: forward the `compat.bodies.file` system property.
- **`.teamcity/settings.kts`**: id17/id18 `export COMPAT_BODIES_FILE=…/trace-data/latest-bodies.ndjson` (**fail-hard** if the sidecar is missing); id16 gains a `COMPAT_BODIES_FILE_RELATIVE` prompt and runs the new test alongside trace replay.

## Guards against vacuous-green (Stage-2 review)

- **Provable execution**: every replayed body is logged via `ExecutionLogger`, so the reporter's proof-of-execution guard counts them and a silently-skipped replay cannot masquerade as a clean run.
- **Fail-hard sidecar** in id17/id18: a missing `latest-bodies.ndjson` (broken checkout) fails the build instead of silently skipping.
- **Distinct-URL guard**: identical baseline/candidate URLs fail-hard (would otherwise be byte-identical → vacuous green).
- **Baseline transport fail-hard** (`status==0`) and **empty-corpus `assumeTrue`** skip.

## Validation

- `:components-registry-compat-test:unitTest` — 12/12 `RealBodyReplayParseTest` green.
- `:components-registry-compat-test:build` — compiles + publication-valid; URL-gated `test` self-skips without stands.
- Two-stage review: Sonnet (correctness) + Opus (adversarial). All FIX findings applied. `RawHttp.postJson` confirmed to serialise a `JsonNode` without double-encoding; `COMPAT_BODIES_FILE` confirmed to reach the test JVM via the build.gradle env→sysprop pass-through.
- Live stand smoke happens on the first `[1.7]` run (booting two stands locally is out of band).

## Intended behaviour

Enabling real bodies in the **auto** `[1.7]`/`[1.8]` chain means a fresh prod-body divergence not in `known-deltas` will **RED** the build — that is the signal we want; triage as a regression or add a reviewed known-delta. id18 uses the empty `known-deltas-git.json`, so any git-mode body diff REDs by design.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -734,7 +734,7 @@ object id17CompatLocalStandManual : BuildType({
                 export COMPAT_VERSIONS_FILE="${'$'}VERSIONS_FILE_PATH"
                 export COMPAT_FULL="%COMPAT_FULL%"
                 export COMPAT_PARALLELISM="%COMPAT_PARALLELISM%"
-                # Real captured POST/PUT bodies (post-bodies.ndjson) replayed by
+                # Real captured POST bodies (post-bodies.ndjson) replayed by
                 # RealBodyReplayCompatTest. Fail-FAST on a missing sidecar (Stage-2
                 # review): the trace repo's master carries latest-bodies.ndjson, so
                 # an absent file means the CrsCompatTrace checkout is broken. Fail-soft
@@ -985,7 +985,7 @@ object id18CompatLocalStandGitModeAuto : BuildType({
                 export COMPAT_VERSIONS_FILE="${'$'}VERSIONS_FILE_PATH"
                 export COMPAT_FULL="%COMPAT_FULL%"
                 export COMPAT_PARALLELISM="%COMPAT_PARALLELISM%"
-                # Real captured POST/PUT bodies (post-bodies.ndjson) replayed by
+                # Real captured POST bodies (post-bodies.ndjson) replayed by
                 # RealBodyReplayCompatTest. Fail-FAST on a missing sidecar (Stage-2
                 # review): the trace repo's master carries latest-bodies.ndjson, so
                 # an absent file means the CrsCompatTrace checkout is broken. Fail-soft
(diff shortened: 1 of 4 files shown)


pgorbachev:
Addressed all review findings in e6686aee:

- **[P1] empty/corrupt sidecar** — `assumeTrue(entries.isNotEmpty())` → `check(...)`. Once `compat.bodies.file` is *configured*, an unreadable or zero-replayable-body sidecar now **hard-fails** the build instead of skipping. Rationale: the TC layer only fail-fasts on file *existence*, and the rest of the compat suite keeps the run's execution count non-zero, so a silent skip would leave the auto [1.7]/[1.8] chain green with the new coverage absent. `assumeTrue`-skip is reserved for the *not-configured* (path unset) case; the in-JVM parse (file readable + ≥1 replayable POST body) is the authoritative parseability gate.
- **POST-only wording** (KDoc + id17/id18 comments) — PUT-with-body is parsed but not replayed; the captured corpus is POST-only.
- **`f.get(90s)` submission-order budget** — documented as a known limitation (this corpus is ~hundreds of bodies, far under the 40k trace; deliberately mirrors TraceReplayCompatTest); a shared ExecutorCompletionService / per-submission-deadline fix is tracked there.

Verified: `:components-registry-compat-test:unitTest` 12/12 green, module compiles clean.

PR #316, opened 2026-06-02
fix(v4): reject containment between field-override ranges (disjoint-only)

pgorbachev:
## Bug

`validateFieldOverrideRange` rejected only **partial overlap** and **exact/semantic duplicates** between two overrides on the same attribute; **strict containment was allowed**. So `build.javaVersion` could hold both:

```
[1.0,2.0] → 17
[0,3.0]   → 18
```

Version `1.5` matches **both** overrides → the resolved value is ambiguous. Reported from the Portal UI.

This was also internally inconsistent with schema-spec §3.5, which allowed containment yet claimed "at most one override per attribute matches V" — containment lets a version match two.

## Fix — disjoint-only

Field-override ranges on one attribute must be **disjoint**. Once `isIntersect` is true it's a conflict (partial overlap, strict containment, and equality all rejected); the semantic-equal case keeps its distinct "duplicate" message.

- Removed the `if (aContainsB || bContainsA) continue` that allowed containment.
- **Composite-sibling gap closed for free:** since rejection now triggers on intersection alone, the containment decision for legacy composite siblings is no longer needed — `isIntersect` (which handles composites correctly) suffices — so a simple new range intersecting a composite sibling is now rejected too (previously deferred/allowed). Composite *new* ranges are still rejected up front.
- `schema-spec.md §3.5` updated to "pairwise disjoint per attribute" (removes the self-contradiction noted above).

Mirrors the Portal client-side change (octopus-components-management-portal#67), so a client `false` and a server 400 now agree.

## Tests (`V4WriteValidationTest`)
- Flipped `allows_strictContainment` → `rejects_strictContainment_innerInsideExisting`.
- Added `rejects_strictContainment_newContainsExisting` (the reported `[1.0,2.0]` + `[0,3.0]` case) and `patch_rejects_strictContainment`.
- Added `allows_adjacentHalfOpenRanges` — boundary guard: `[1.0,2.0)` + `[2.0,3.0)` must stay **allowed** (touching-exclusive ≠ intersecting), since `isIntersect` is now the sole gate.
- Disjoint, semantic-equal, and PATCH self-reassign tests unchanged.

`./gradlew :components-registry-service-server:test --tests V4WriteValidationTest` → **BUILD SUCCESSFUL**. Sonnet review: no P0; P1 (the boundary test) and a doc nit addressed.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -1427,10 +1427,12 @@ class ComponentManagementServiceImpl(
         }
     }
 
-    // ─── Field-override range validation (partial-overlap, matches Portal) ───────
+    // ─── Field-override range validation (disjoint-only, matches Portal) ─────────
     //
-    // The Portal applies the same partial-overlap and semantic-equality rules
-    // client-side via `versionRange.ts` (`rangesOverlap`). Keeping the two
+    // The Portal applies the same disjoint-only rule client-side via
+    // `versionRange.ts` (`rangesOverlap` / `classifyRangeConflict`): two
+    // overrides on one attribute conflict if they intersect at all — partial
+    // overlap, strict containment, or semantic equality. Keeping the two
     // implementations in sync is the explicit goal — a client-side `false`
     // should not surprise the user with a server 400 and vice versa.
     //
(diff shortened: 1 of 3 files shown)


pgorbachev:
Addressed both Copilot comments on `schema-spec.md` (§3.5 and the validation bullet): reworded so the two mechanisms are no longer conflated — the **service validator** (`validateFieldOverrideRange`) rejects all intersections including *semantic* equality (whitespace / trailing-zero differences like `[1.0, 2.0)` or `[1,2)`), while the **DB UNIQUE** on `(component_id, version_range, overridden_attribute)` only blocks *exact-text* duplicates. Pushed as `c45fb677`.

PR #317, opened 2026-06-02
fix(MIG-039,MIG-040,MIG-023): find-by-artifacts per-config matching + specificity ranking; find-by-docker-images version-substitution

pgorbachev:
## What

Two schema-v2 `find-by-*` regressions surfaced by the real-body replay (PR #315), fixed to **mirror the V1 resolver**. (Supersedes this PR's first attempt, which matched `distribution.GAV()` — reverted: V1 never consults GAV, so that over-resolved and was version-range-blind. See the held-review thread.)

### MIG-039 — `find-by-artifacts` must gate by configuration version range
`DatabaseComponentRegistryResolver.findComponentByArtifactOrNull` matched only `component_artifact_ids` group/artifact patterns and tie-broke by `artifactPattern.length`, **ignoring version ranges**. An archived/old component whose long `|`-union `artifactId` still matched the name (but whose ranges excluded the version) then won the tie-break → wrong component (live repro: an artifact at version `11.1.157` resolved to an archived component whose ranges are all `[1.0.x,…)` instead of the active one).

**Fix:** mirror `EscrowModuleConfigMatcher` — a group/artifact pattern match is necessary but not sufficient; the artifact version must also fall within one of the component's configuration version ranges (`versionInRange`, reusing `versionRangeFactory`/`numericVersionFactory`). Specificity/length tie-break kept only as a fallback after the gate.

### MIG-040 — `find-by-docker-images` must version-substitute the distribution
`findConfigurationByDockerImage` resolved via `toResolvedEscrowModuleConfig`, whose `distribution.docker()` lacks the version tag (`composeDockerCsv` emits `img`, not `img:<version>`, when flavor is null), so the `"$imageName:$imageTag"` membership check never matched → empty result where V1 returns the image (live repro: 22 vs 0).

**Fix:** resolve via `getResolvedComponentDefinition` so `EscrowConfigurationLoader.calculateDistribution` substitutes the version, exactly like the `/versions/{v}/distribution` endpoint.

## TDD / verification
- `DatabaseComponentRegistryResolverFindByArtifactTest` — **RED-then-GREEN proven**: 2/3 cases fail on the pre-fix resolver (wrong-component via length tie-break + resolves out-of-range), all pass with the gate. Range + routing suites stay green.
- MIG-040's behavioural proof is the next `[1.7]` real-body replay (the escrow-config resolution path isn't practically unit-mockable).
- Diagnosis from a live baseline-V1 + candidate-DB-mode repro (both stands, 953/953 migrated).

## Out of scope — MIG-041 (follow-up)
The same repro found a second, narrower defect: a component whose component-level `artifactId` is a CSV (`a,b`) migrates only the first version-range block's token, so the reverse lookup can't map the dropped token. **It's deferred to a focused follow-up** because `component_artifact_ids` is also the `/maven-artifacts` fallback (which expects the component *default* artifactId), so a safe fix needs the true component default + an integration test over **both** endpoints. Registered as MIG-041 in `requirements-migration.md`.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/DatabaseComponentRegistryResolver.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/DatabaseComponentRegistryResolver.kt
@@ -10,7 +10,6 @@ import org.octopusden.octopus.components.registry.core.dto.Image
 import org.octopusden.octopus.components.registry.core.dto.VersionedComponent
 import org.octopusden.octopus.components.registry.core.exceptions.NotFoundException
 import org.octopusden.octopus.components.registry.server.config.ConditionalOnDatabaseEnabled
-import org.octopusden.octopus.components.registry.server.entity.ComponentArtifactIdEntity
 import org.octopusden.octopus.components.registry.server.entity.ComponentConfigurationEntity
 import org.octopusden.octopus.components.registry.server.entity.ComponentEntity
 import org.octopusden.octopus.components.registry.server.mapper.MarkerAttributes
@@ -20,8 +19,6 @@ import org.octopusden.octopus.components.registry.server.repository.ComponentRep
 import org.octopusden.octopus.components.registry.server.repository.DependencyMappingRepository
 import org.octopusden.octopus.components.registry.server.service.ComponentRegistryResolver
 import org.octopusden.octopus.components.registry.server.util.formatVersion
-import org.octopusden.octopus.components.registry.server.util.parseMavenGavEntry
-import org.octopusden.octopus.components.registry.server.util.splitCsv
 import org.octopusden.octopus.escrow.MavenArtifactMatcher
 import org.octopusden.octopus.escrow.ModelConfigPostProcessor
 import org.octopusden.octopus.escrow.config.JiraComponentVersionRange
@@ -265,13 +262,28 @@ class DatabaseComponentRegistryResolver(
 
     override fun findComponentsByArtifact(artifacts: Set<ArtifactDependency>): Map<ArtifactDependency, VersionedComponent?> {
         log.debug("Find components by artifacts: {}", artifacts)
-        return artifacts.associateWith { artifact -> findComponentByArtifactOrNull(artifact) }
+        // Build per-component per-range effective (group, artifact) ONCE, then match each artifact
+        // — avoids re-reading all components and reconstructing their escrow view per artifact.
+        val perComponent = loadPerComponentArtifactParameters()
+        return artifacts.associateWith { artifact -> matchArtifact(artifact, perComponent) }
     }
 
     override fun getMavenArtifactParameters(component: String): Map<String, ComponentArtifactConfiguration> {
         val componentEntity =
             componentRepository.findByComponentKey(component)
                 ?: throw NotFoundException("Component '$component' is not found")
+        return mavenArtifactParametersFor(componentEntity)
+    }
+
+    /**
+     * Per-range effective `(groupPattern, artifactPattern)` for a component — marker-aware
+     * (`GROUP_ARTIFACT_PATTERN` per-range overrides) with the component-level `artifactIds`
+     * fallback. Shared by `/maven-artifacts` and (MIG-039) `find-by-artifacts` so both use
+     * identical per-range semantics; keyed by version-range string.
+     */
+    private fun mavenArtifactParametersFor(
+        componentEntity: ComponentEntity,
+    ): Map<String, ComponentArtifactConfiguration> {
 
         // Component-level fallback: the top-level groupId/artifactId rows imported from DSL.
         // For components whose DSL omits an explicit `artifactId` line, the import path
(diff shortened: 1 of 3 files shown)


components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/DatabaseComponentRegistryResolver.kt
```diff
@@ -539,10 +540,30 @@ class DatabaseComponentRegistryResolver(
...
                 .flatMap { component ->
-                    component.artifactIds.mapNotNull { artifactIdEntity ->
-                        toArtifactMatchOrNull(artifactIdEntity, artifact)
-                    }
+                    val byArtifactId =
+                        component.artifactIds.mapNotNull { artifactIdEntity ->
+                            toArtifactMatchOrNull(artifactIdEntity, artifact)
+                        }
+                    // MIG-039: component_artifact_ids is populated only from DSL
+                    // groupId/artifactId. Coordinates declared via `distribution { gav = … }`
```
pgorbachev: [P1] This scan is too broad in two ways. First, it reads every configuration `mavenArtifacts` row, including `DISTRIBUTION_MAVEN` rows parsed from `distribution { gav = ... }`; the legacy `find-by-artifacts` path matches only `EscrowModuleConfig.groupIdPattern` / `artifactIdPattern`, not `distribution.GAV()`, so DB mode can now resolve distribution-only artifacts that V1 would not resolve. Second, these rows belong to `ComponentConfigurationEntity` ranges, but the lookup ignores the owning row versionRange; the legacy matcher only accepts a config if its version range contains the artifact version, so a range-only artifact can resolve for versions outside that range.

pgorbachev (commented):
Review findings from the local pass. Leaving these as comments rather than changing the PR review state.

pgorbachev:
Converting to **draft / hold** — the P1 is correct and verified against the V1 matcher.

`EscrowModuleConfigMatcher.match` (V1's `find-by-artifacts` path) matches **only** `moduleConfig.groupIdPattern`/`artifactIdPattern` **and** requires `versionRange.containsVersion(artifact.version)`; `ModuleByArtifactResolver` then requires exactly one matching config. It never consults `distribution.GAV()`.

So this branch's GAV-matching approach is wrong-direction: it over-resolves (GAV-only artifacts V1 wouldn't) and is version-range-blind. It only zeroed the replayed-set diffs because those artifacts coincided.

**Plan:** hold this PR; do a live repro (boot baseline V1 + candidate DB-mode stands) to find the *true* reason the DB matcher diverges from V1's per-range (groupId, artifactId)+version-range logic — likely `component_artifact_ids` being component-level / missing per-range/inherited patterns. Same repro pins MIG-040 (docker). Then rework the fix to faithfully mirror `EscrowModuleConfigMatcher` (and fold in the Copilot nits: MIG-039 in test names + `@DisplayName`, `**Test method:**` doc field, specificity test, N+1 fetch-join).

pgorbachev:
Reworked per the [P1] — the version-range gate was only *partially* faithful (it gated by 'any range contains the version' but matched **component-level** `artifactIds`, decoupling range from pattern).

**Fix:** `find-by-artifacts` now mirrors V1's `EscrowModuleConfigMatcher` **per config** — for each component, a version range whose **effective** `(groupPattern, artifactPattern)` matches the artifact **and** that contains the version. It reuses the same marker-aware per-range resolution as `/maven-artifacts` (extracted `mavenArtifactParametersFor`), so:
- per-range `artifactId` overrides resolve on the **same** config as V1 — a superseded artifact no longer resolves for a version in the overriding range; a per-range `new` artifact resolves in its range (the exact cases you raised: `old:2.1`→null, `new:2.1`→resolves);
- archived/old out-of-range ranges still don't pollute resolution.
Per-call cache so the batch endpoint reads all components once.

**Tests:** `DatabaseComponentRegistryResolverFindByArtifactTest` now covers version-range gating, the per-range override (resolve + supersede), and the null guards — RED-then-GREEN. Broader `DatabaseComponentRegistryResolverTest` / `ComponentRoutingResolver*` / `MavenArtifactsRange*` suites stay green.

**[P3]** added the `**Test method:**` field to MIG-039 in `requirements-migration.md`.

(MIG-041 — importer dropping component-level `artifactId` CSV tokens — remains the documented follow-up: `component_artifact_ids` is also the `/maven-artifacts` fallback, so it needs the true component default + an integration test over both endpoints.)

pgorbachev:
Fixed the [P1] MIG-023 regression. The per-config matcher returned the **first** in-range hit, dropping the `artifactSpecificity` ranking — so a generic mapping could beat a more specific one depending on `findAll()` order (and the body's 'specificity fallback kept' claim was untrue in the code).

**Fix:** `matchArtifact` now collects **all** in-range matches across components and ranks them by `artifactSpecificity` (exact > `|`-union > `*` wildcard, then pattern length) after the range gate — so the specific/version-aware mapping wins regardless of `findAll()` order, per MIG-023. (V1 would conflict on >1; a batch endpoint instead picks the most specific.)

**Guard test added** (RED-then-GREEN): `MIG-039 prefers the most specific in-range mapping over a generic one listed first` — generic component returned first, exact second → resolves the **exact** one. Fails on the first-hit code, passes now. Full find-by-artifact (6) + `DatabaseComponentRegistryResolverTest` + `ComponentRoutingResolver*` + `MavenArtifactsRange*` suites green.

pgorbachev:
Fixed the [P1] catch-all-regex MIG-023 gap. `artifactSpecificity` only downgraded the **literal** `*`, so the inherited default `ANY_ARTIFACT` regex (`.*` / `[\\w-\\.]+`, written verbatim for components without an explicit `artifactId`) fell into `else -> 2` (specific) and beat a concrete `core|api` / `core,api` union (`1`).

**Fix:** catch-all is now detected **behaviorally** — `isCatchAllArtifactPattern` returns true if the pattern matches an arbitrary probe id (`x-any-artifact-probe-0.0`). That matches `.*`, `[\\w-\\.]+`, and literal `*`, but not concrete names or `|`/`,` unions — so it covers whichever default regex the DSL uses, not just `*`. Catch-all → specificity 0, below any concrete mapping.

**Guard test (RED→GREEN):** `MIG-039 ranks the default catch-all regex below a concrete union mapping` — generic `[\\w-\\.]+` first, concrete `core|api` second → resolves the **concrete** one. Fails on the prior code, passes now. All 7 find-by-artifact tests + `DatabaseComponentRegistryResolverTest` / `ComponentRoutingResolver*` / `MavenArtifactsRange*` green.

pgorbachev:
Fixed the [P1] dot-less catch-all gap. The probe `x-any-artifact-probe-0.0` contained a dot, so the `[\\w-]+` default form (no dot) didn't match it and escaped catch-all detection → scored 2 and could beat a concrete union (1).

**Fix:** the probe is now **pure lowercase letters** — `xanyartifactprobe` — with no dot, dash, digit or underscore. That string is matched by every `[\\w…]`-family default (`[\\w-]+`, `[\\w-\\.]+`, `.*`, `\\w+`, `[a-z]+`) while still never matching a concrete name or `|`/`,` union. A probe with a dot misses `[\\w-]+`; one with a dash would miss `\\w+` — pure lowercase covers all of them.

**Guard test (RED→GREEN):** `MIG-039 ranks the dot-less default catch-all regex below a concrete union mapping` — generic `[\\w-]+` first, concrete `core|api` second → resolves the concrete one. RED on the dotted probe, GREEN now. 8 find-by-artifact cases + `DatabaseComponentRegistryResolverTest` / `ComponentRoutingResolver*` / `MavenArtifactsRange*` all green.

PR #318, opened 2026-06-02
feat(security): add ADMIN_DATA permission, mapped to ROLE_ADMIN

pgorbachev:
## What

Adds an admin-tier **`ADMIN_DATA`** permission to gate power-user / escape-hatch editing surfaces — specifically the Portal **Field-Overrides** tab (add/edit/delete, incl. marker editing). Distinct from `IMPORT_DATA` (bulk import).

- `PermissionEvaluator`: `const ADMIN_DATA` + `canAdminData()` helper (mirrors `canImport()` etc.).
- `octopus-security.roles`: `ADMIN_DATA` added to `ROLE_ADMIN` in **all** role-config files — server `application.yml` / `application-dev.yml`, the **OKD deploy config** (`components-registry-automation/data/components-registry-service.yaml`, injected as `application-test` into the running pod), and the server/client/light-client test resources.

## Why no Keycloak change

Role→permission mapping is **CRS config** (`octopus-security.roles`), not Keycloak. Keycloak only assigns *roles* (realm-role `ADMIN` → `ROLE_ADMIN`); CRS resolves roles→permissions and `/auth/me` returns them. So anyone already holding the `ADMIN` realm-role automatically gains `ADMIN_DATA` once this deploys — no IdP/DevOps coordination needed.

## Safety

Granting `ADMIN_DATA` to `ROLE_ADMIN` is **inert** on its own — no `@PreAuthorize` keys off it yet (the gate is currently the Portal UI). It changes zero server behavior; it just makes the permission available. Added only to `ROLE_ADMIN` (not VIEWER/EDITOR/ANONYMOUS).

**Deploy ordering:** this should deploy **before** the paired Portal change (octopus-components-management-portal#NN), so admins carry `ADMIN_DATA` before the UI starts gating on it.

## Verification
- TDD: `UserInfoConverterIntegrationTest` now asserts an admin's `/auth/me` permissions include `ADMIN_DATA`. `./gradlew :components-registry-service-server:test --tests UserInfoConverterIntegrationTest` → BUILD SUCCESSFUL.
- Sonnet review: the P1 (4 role-config files initially missed, incl. the OKD deploy config) is fixed in this revision; all 7 now carry it.

### DIFF
--- a/components-registry-automation/data/components-registry-service.yaml
+++ b/components-registry-automation/data/components-registry-service.yaml
@@ -57,6 +57,7 @@ octopus-security:
       - DELETE_COMPONENTS
       - IMPORT_DATA
       - ACCESS_AUDIT
+      - ADMIN_DATA
 
 logging:
   level:
--- a/components-registry-service-client/src/test/resources/application-common.yml
+++ b/components-registry-service-client/src/test/resources/application-common.yml
@@ -55,3 +55,4 @@ octopus-security:
       - DELETE_COMPONENTS
       - IMPORT_DATA
       - ACCESS_AUDIT
+      - ADMIN_DATA
--- a/components-registry-service-client/src/test/resources/application-v3.yml
+++ b/components-registry-service-client/src/test/resources/application-v3.yml
@@ -53,3 +53,4 @@ octopus-security:
       - DELETE_COMPONENTS
       - IMPORT_DATA
       - ACCESS_AUDIT
+      - ADMIN_DATA
--- a/components-registry-service-light-client/src/test/resources/application-common.yml
+++ b/components-registry-service-light-client/src/test/resources/application-common.yml
@@ -52,3 +52,4 @@ octopus-security:
       - DELETE_COMPONENTS
       - IMPORT_DATA
       - ACCESS_AUDIT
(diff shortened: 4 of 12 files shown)


pgorbachev:
Addressed the P2 doc finding (pushed as `c3cccd6d`):

- **ADR-004** — permission count 7 → 8; added the `ADMIN_DATA` row to the permissions table and to the `ROLE_ADMIN` entry in the role-map YAML example.
- **functional-spec.md** — permission list 7 → 8, with `ADMIN_DATA`.
- **keycloak-setup.md** — added `ADMIN_DATA` to the `ROLE_ADMIN` row, **plus** a new operator note: config-overlay *list* properties REPLACE (not append), so when a new permission is added it must also be added to `ROLE_ADMIN` in every `service-config` overlay (`<env>/service-config: components-registry-service.yml`) — otherwise admins keep `ROLE_ADMIN` but silently lose the capability. Verify post-deploy via an admin's `/auth/me` or the startup role→permission log.

The 4 other docs that mention `ROLE_ADMIN`/`IMPORT_DATA` only do so incidentally (no canonical permission-set enumeration), so they were left unchanged.

PR #325, opened 2026-06-03
feat(compat): [1.7]/[1.8] replay 30k trace slice + full business-endpoint coverage (closes #324)

pgorbachev:
## What

Make the **[1.7]/[1.8] AUTO compat gate** (TeamCity id17 db-mode / id18 git-mode) replay a **30,000-request** slice of the production trace on every build, on top of the existing ~1,716 real-body replays — and guarantee **every business endpoint** is exercised. The full `latest-top.txt` (~75k unique tuples now) stays unlimited for the manual id16 tier.

## Changes

- **`TraceReplayCompatTest`** — `compat.trace.limit` / `COMPAT_TRACE_LIMIT` caps the replay to the top-N highest-traffic tuples (`parseTraceLimit` / `applyTraceLimit`; the trace is rank-ordered, so take(N) = the busiest N). After capping, `ensureEndpointCoverage` re-adds the busiest representative of every business-endpoint **template** the cap dropped. `endpointTemplate` collapses `{component}`/`{version}`/`{project}` and preserves `find*` actions (incl. the legacy camelCase `findByArtifacts`).
- **`build.gradle`** — forward `compat.trace.limit`.
- **`.teamcity/settings.kts`** — id17 + id18 export `COMPAT_TRACE_FILE=$TRACE_DATA_DIR/latest-top.txt` + `COMPAT_TRACE_LIMIT=30000` (fail-fast if the trace file is missing). `RealBodyReplayCompatTest` unchanged.
- **Dedicated tests for zero-trace endpoints (closes #324):** `CopyrightCompatTest` (`GET .../copyright` — binary, byte-compare with **content redaction**), `DistributionCompatTest` (component-level v1/v2 + per-version), `ComponentsListV1CompatTest` (v1 list + detail).
- **`TraceReplayParsingTest`** — unit tests for the four helpers (RED→GREEN coverage union; `findByArtifacts` templating regression guard).
- **TD-008 doc** — limit knob, replay tiers, and the (now-closed) endpoint-coverage map.

## Coverage scope

Every business endpoint is now exercised by one of: the **30k trace slice** (frequency-weighted, with the coverage-union backstop for any trace-present endpoint the cap dropped), the **real-body replay** (find-by-* POSTs), or a **dedicated `*CompatTest`** (copyright / component-distribution / v1 list — the three endpoints with zero prod-trace hits, **#324**).

## Verification

- `:components-registry-compat-test:unitTest` green (12 tests); `compileTestKotlin` green for the new suites.
[...]

PR #326, opened 2026-06-03
fix(compat): collision-free jira-ranges sort key + id17/id18 green-skip on main

pgorbachev:
## What

Two fixes for the v3 `[1.7]/[1.8]` compat-gate fallout (builds **3778** = exit 127, **3779** = 50 diffs):

### 1. Collision-free `jira-component-version-ranges` sort key (fixes 44/50 diffs on 3779)
`RawArraySorters` sorts `/projects/{p}/jira-component-version-ranges` (and the `/common/` variant) by `(componentName, versionRange)` — but that pair is **not unique** within a project's response: several entries share it yet differ in `distribution` / `vcsSettings` / `component.displayName`. The colliding key let the stable sort keep each stand's arbitrary Set-iteration order → positional misalignment → **44 false `STRUCTURAL_DIFF`s** (the `displayName`/`externalRegistry`/`versionControlSystemRoots` NULL↔STRING / 1↔0 flips, and the `distribution.GAV` element landing at index 6 on one stand, 8 on the other).

**Fix:** append a **canonical** (recursively field-name-sorted) serialization of the whole element as a tie-breaker → collision-free + deterministic on both stands. It's **non-masking** — a genuine per-element diff still surfaces (the existing anti-regression tests prove real value/size/shape diffs still fire). RED→GREEN unit test added.

### 2. id17/id18 green-skip when the compat infra is absent (fixes the 3778 exit 127)
`[1.7]/[1.8]` auto-fire on `[1.0]` success for any branch. When a build resolves to `main` (no compat infra) — or a **deleted feature branch falls back to the default branch** (exactly what happened to 3778 after #325 merged + the branch was deleted) — `scripts/local-stands/teamcity-run.sh` is absent and the step exited **127**. Now it green-skips with a clear status, mirroring the existing id15/id16 main-absent guard.

## Not included
The remaining **6/50 diffs** are a real status divergence — `GET /rest/api/2/components/{c}/versions/{v}/distribution` returns baseline **404** vs candidate **200** — under separate investigation (over-resolution vs intended). Not a comparator issue.

## Verification
- `:components-registry-compat-test:unitTest` green — new collision test + all anti-regression non-masking guards.
- `.teamcity` `mvn compile` green (DSL).

Targets `v3`.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -643,6 +643,17 @@ object id17CompatLocalStandManual : BuildType({
             // so there is no port / Postgres collision between them.
             scriptContent = """
                 set -euo pipefail
+                # Green-skip when the compat-test infra isn't in this checkout. [1.7]/[1.8]
+                # auto-fire on [1.0] success for ANY branch; if a build resolves to `main`
+                # (the legacy V1 trunk) — or a deleted feature branch falls back to the
+                # default branch — `scripts/local-stands/` is absent and the wrapper below
+                # would exit 127. Mirror id15/id16: succeed with a clear status, not a red 127.
+                if [ ! -f scripts/local-stands/teamcity-run.sh ]; then
+                    echo "::: scripts/local-stands/teamcity-run.sh not present in this checkout."
+                    echo "::: [1.7] is only meaningful on v3-family branches that carry the compat-test infra."
+                    echo "##teamcity[buildStatus status='SUCCESS' text='Skipped: compat-test infra absent on this branch (likely main); run from v3 family instead.']"
+                    exit 0
+                fi
                 WORK_DIR="/tmp/crs-id17-%teamcity.build.id%"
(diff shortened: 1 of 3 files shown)


pgorbachev:
Addressed in 5dd08994: renamed `canonicalJson` -> `canonicalSortKey` (it's a deterministic sort key, not a JSON document) and quoted the field names so the concatenation is unambiguous (a value containing `:` can no longer collide with a field-name/value boundary). Behaviour is unchanged for the real DTOs; `RawArraySortersTest` stays green (collision test + anti-regression guards).

FWIW build 3783 (a different branch, without this fix) re-ran with **46** jira-component-version-ranges STRUCTURAL_DIFFs across ANCS/CARDS/**WWEB** — vs 44 across ANCS/CARDS on 3779. The count and the set of projects vary run-to-run, which is the non-determinism this PR's collision-free key eliminates.

PR #336, opened 2026-06-04
fix(mapper): gate toResolvedEscrowModuleConfig on base versionRange (MIG-042)

### LINKED ISSUE #329: MIG-042: v3 over-resolves per-version distribution for out-of-range versions (V1 404 vs v3 200)
Surfaced by the v3 [1.7] compat gate (build 3779). `GET /rest/api/2/components/{component}/versions/{version}/distribution` returns **404 on the V1 baseline** but **200 on the v3 (DB-mode) candidate** for versions that fall **outside the component's configured version-ranges**.

## Root cause
The 404/200 is entirely `componentRegistryResolver.getResolvedComponentDefinition(component, version)` (null → 404; `BaseComponentController.getComponentVersionDistribution`).

- **V1** `EscrowConfigurationLoader.resolveComponentConfiguration` filters **all** module configs by `versionRange.containsVersion(version)` and returns **null** when the version is outside every range → 404.
- **v3** `ComponentEntity.toResolvedEscrowModuleConfig` (server `mapper/EntityMappers.kt`) returns null only when there is no `BASE` row; it then returns that BASE config for **any** parseable version — only the *overrides* are range-gated, the **BASE is not**. → returns a config → 200.

So v3 over-resolves: it serves the base distribution for versions V1 rejects as out-of-range.

## Proposed fix
Gate the BASE in `toResolvedEscrowModuleConfig` by `versionRange.containsVersion(version)` (return null when the version is outside the component's configured range(s)), mirroring V1. TDD: out-of-range version → null; in-range still resolves; multi-range components still work.

## Blast radius (why this needs care)
`getResolvedComponentDefinition` also feeds `find-by-docker-images` and the "component as code" view — a gating change must be regression-tested against those, not just the distribution endpoint.

## Severity / context
- **Intermittent**: only surfaces when the compat version-sampler picks an out-of-range version (absent from build 3783's smaller sample).
- **Low-traffic**: this endpoint has ~zero production traffic and carries a `// todo - consider removing the whole endpoint` in `BaseComponentController` (version isn't meaningful for component-level distribution).

Disposition options: (a) fix as above; (b) accept + a documented known-delta with a pinning test; (c) track only. Pending decision.

pgorbachev:
## Problem

`GET /rest/api/2/components/{component}/versions/{version}/distribution` returned HTTP **200** on the v3 (DB-mode) candidate for versions outside the component's configured version range, while the V1 baseline returned **404**. Surfaced by the [1.7] compat gate: **50 active STATUS_CODE_DIFF** divergences (build 3779), **46** on build 3783.

Fixes #329.

## Root cause

`ComponentEntity.toResolvedEscrowModuleConfig` in `EntityMappers.kt` returned null only when no BASE row exists. The override rows were range-gated, but the BASE was not — so any parseable version got a config back. V1's `EscrowConfigurationLoader.resolveComponentConfiguration` gates everything through `versionRange.containsVersion(version)`.

## Fix

After parsing `numericVersion`, check `versionRangeFactory.create(base.versionRange).containsVersion(numericVersion)` and return null when the version falls outside the base range. Unparseable ranges fall through conservatively.

## Tests

Added 4 test cases `(9a–9d MIG-042)` to `DatabaseComponentRegistryResolverTest`:
- **(9a)** in-range version → non-null (no regression)
- **(9b)** out-of-range version → null (the core fix)
- **(9c)** ALL_VERSIONS base → any version still resolves (no regression)
- **(9d)** synthetic base (MIG-029) + bounded range → out-of-range → null

### DIFF
--- a/components-registry-service-client/src/test/resources/testcontainers.properties
+++ b/components-registry-service-client/src/test/resources/testcontainers.properties
@@ -1,0 +1,5 @@
+# Disable Ryuk resource reaper: CI agents deny the 'statfs /var/run/docker.sock'
+# syscall that Ryuk uses to verify the Docker socket is a bind-mount, which causes
+# initializationError before any test runs.  With Ryuk disabled, Testcontainers
+# still starts PostgreSQL containers normally; they are stopped when the JVM exits.
+ryuk.disabled=true
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/CommonControllerV2.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/CommonControllerV2.kt
@@ -26,11 +26,11 @@ class CommonControllerV2(
     // todo - consider removing the whole endpoint or just version specific fields, like docker,
     //  because version is not provided in this context
     @GetMapping("jira-component-version-ranges", produces = [MediaType.APPLICATION_JSON_VALUE])
-    fun getAllJiraComponentVersionRanges(): Set<JiraComponentVersionRangeDTO> =
+    fun getAllJiraComponentVersionRanges(): List<JiraComponentVersionRangeDTO> =
         componentRegistryResolver
             .getAllJiraComponentVersionRanges()
             .map { it.toDTO() }
-            .toSet()
+            .sortedWith(compareBy({ it.componentName }, { it.versionRange }))
 
     @GetMapping("dependency-aliases", produces = [MediaType.APPLICATION_JSON_VALUE])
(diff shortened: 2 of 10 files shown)


pgorbachev:
Superseded by #342. The MIG-042 version-range gate was re-derived there as a configured-range UNION gate (covers multi-block gaps + RANGE_PRESENCE rows — cases this branch's base-range approach didn't), under the diff-of-diffs NEW==0 process; the Ryuk CI fix (b1183b8b) was cherry-picked into #342. The full [1.7] sweep reached 0 active without this branch's remaining changes (VAL-011 empty-tools-marker, RES-013 ALL_VERSIONS guard, externalRegistry-null-on-marker), so they correspond to no real diff on the current 30k trace. Closing; branch deleted.

Issue #82, opened 2025-05-02
Docker images - tag format


fishinitself:
As a solution to the problem of finding a match between a Docker image and its components, conventions for the form of a Docker image tag are suggested.

The following format is proposed for the image tag `version[-suffix]` where the suffix is ​​a word containing at least one letter. 
When the part with version is always predefined and does not mentioned. The part with suffix is optional.
The tag cannot contain any expressions, including SpEL 

Required steps - CR validation and CR migration

PR #350, opened 2026-06-09
feat: generate OpenAPI v4 spec, gate drift, publish as CI artifact (TD-003)

pgorbachev:
## What & why

CRS **TD-003**. The Portal SPA hand-maintains its v4 wire-contract types; [ADR-012](docs/registry/adr/012-portal-architecture.md) named OpenAPI generation as the mitigation for the API/UI separate-repo split. This wires the **CRS side** (Phase A of a 3-phase plan): generate a machine-readable v4 spec, gate it for drift, and publish it so the Portal can derive `schema.d.ts` from it (Portal issue #82, Phase C).

Scope is the **v4** surface only — v1/v2/v3 stay hand-shaped Feign contracts.

## How

- **`OpenApiV4Config`** — `GroupedOpenApi("v4")` matching the v4 controllers + `/auth/**`, served at `GET /v3/api-docs/v4` (already `permitAll`). `info` is scoped to the group (a global bean would relabel the ungrouped `/v3/api-docs`); `info.version` is the **constant** `"4"`, not the build version, so the committed spec only changes on real API-surface changes.
- **`OpenApiV4SpecTest`** — `@SpringBootTest`+MockMvc reusing the `BasicFunctionalitySmokeTest` H2/`smoke` context (shared test-context cache; no Keycloak/Postgres). Captures the doc, **canonicalizes** (recursive object-key sort + fixed 2-space `\n` pretty-print + trailing newline → byte-stable), writes `build/openapi/v4.json`, and asserts byte-equality vs the committed `src/main/resources/openapi/v4.json`. Also asserts the six v4 groups are present and **no** v1/v2/v3 paths. UNtagged → runs under `test → check → build` (the `[1.0]` gate), no new gate task.
- **`generateOpenApiDocs`** — `./gradlew :components-registry-service-server:generateOpenApiDocs` refreshes the committed spec after an intended change. Two tasks (`generateOpenApiDocsTest` Test → `generateOpenApiDocs` Copy via `dependsOn`) so a failing regeneration never copies a stale spec. Separate output dir from the gate run so the two Test tasks can't race.
- **`application.yml`** — `springdoc.writer-with-order-by-keys: true`.
- **TeamCity `[1.0]`** — publishes `components-registry-service-server/build/openapi/v4.json` under the `openapi` artifact path for Portal CI to pull.
- **docs** — TD-003 status/AC update (fixes the stale `/audit-log` → `/audit` path and the "no springdoc dep" line), new `api-changelog.md`.

## Verification

- Full `./gradlew build` green (drift gate runs in the server `test`).
- `generateOpenApiDocs` run twice → byte-identical `v4.json` (same shasum).
- RED demo: tampering the committed spec fails the gate; `generateOpenApiDocs` restores it green.
- Generated spec covers all six groups, no v1/v2/v3 paths, `info.version="4"`.

## Follow-ups (not in this PR)

- **Phase B** (CRS): structural `@Schema` metadata on the v4 DTOs.
- **Phase C** (Portal #82): vendor the published spec, regenerate `schema.d.ts`, migrate `types.ts`.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -137,6 +137,7 @@ object id10CompileUtAuto : BuildType({
         %ARTIFACT_PATH%
         **/reports
         **/test-results
+        components-registry-service-server/build/openapi/v4.json => openapi
     """.trimIndent()
 
     params {
--- a/components-registry-service-server/build.gradle
+++ b/components-registry-service-server/build.gradle
@@ -165,8 +165,11 @@ processIntegrationTestResources {
 def openApiGateDir = layout.buildDirectory.dir('openapi').get().asFile
 def openApiRefreshDir = layout.buildDirectory.dir('openapi-refresh').get().asFile
 test {
+    // Only set the output dir — the gating `test` run must ALWAYS assert (never refresh), so
+    // `openapi.refresh` is deliberately NOT wired from a -P here. `-Popenapi.refresh=true` would
+    // otherwise make `./gradlew build` silently skip the drift assertion. Refresh is the dedicated
+    // `generateOpenApiDocs`/`generateOpenApiDocsTest` path only.
     systemProperty 'openapi.outputDir', openApiGateDir.absolutePath
-    systemProperty 'openapi.refresh', project.findProperty('openapi.refresh') ?: 'false'
 }
 
 // `./gradlew :components-registry-service-server:generateOpenApiDocs` refreshes the committed
--- a/components-registry-service-server/build.gradle
+++ b/components-registry-service-server/build.gradle
@@ -156,6 +156,50 @@ processIntegrationTestResources {
     duplicatesStrategy = DuplicatesStrategy.INCLUDE
 }
 
+// TD-003: OpenApiV4SpecTest captures GET /v3/api-docs/v4 into build/openapi/v4.json and gates
+// drift vs the committed src/main/resources/openapi/v4.json. The test is UNtagged, so it already
+// runs under `test` -> `check` -> `build` (the [1.0] gate) — no extra gate task. Feed it the
+// output dir and the refresh flag (default false: gate; true: refresh the committed copy).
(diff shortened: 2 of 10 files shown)


pgorbachev:
Pushed `5e9d57aa`: fixed an OpenAPI path-template collision — the item-level GET used `/{idOrName}` while PATCH/DELETE use `/{id}`. OpenAPI path identity ignores the parameter name, so the generator emitted two colliding path items for the same `/components/{id}` URL (invalid; `openapi-typescript` could drop/merge one). Now bound to `{id}` via `@PathVariable("id")` while keeping the `idOrName` Kotlin param (GET still resolves by UUID or name). Added a normalized-template uniqueness assertion to `OpenApiV4SpecTest` and refreshed `v4.json` (get/patch/delete now merged under one path item, each keeping its own parameter schema).

pgorbachev:
Addressed review (`88d74aa3`):
- **refresh footgun** — the main `test` task no longer wires `openapi.refresh` from a -P property, so `./gradlew build` always runs the drift assertion; refresh stays the dedicated `generateOpenApiDocs`/`generateOpenApiDocsTest` path.
- **charset robustness** — `OpenApiV4SpecTest` now decodes the api-docs response via `contentAsByteArray` + explicit UTF-8 instead of `contentAsString` (response charset).
- **docs** — `api-changelog.md` moved out of the "Planned" table into a plain note (it now exists); table renders cleanly.
- Spec regenerated byte-identically (ASCII-only); no wire/behaviour change.

On the **`quality/tests-coverage`** red: it's infra, not this change. The failed run shows a JVM `OutOfMemoryError` + a JaCoCo `libinstrument` assertion + `Connection to localhost:32768 refused` (Postgres testcontainer dropped) — no test-assertion failures. The `build` job (which runs the new `OpenApiV4SpecTest` drift gate) passes in ~2m, and the first commit's `tests-coverage` re-run previously went green at 14m48s. A spec-generation + path-variable rename can't OOM the migration DB suite. Re-running.

PR #355, opened 2026-06-10
fix: section-field saves must write an audit row (SYS-053)

pgorbachev:
## Problem

Saving a section of the component form (e.g. Build tab → Maven Version → "Save Build") persisted the value but the **History tab stayed empty**. Other section tabs (VCS / Jira / Escrow / Distribution, child collections) were affected the same way. Component **creation showed up in History normally**.

## Root cause

The audit snapshots passed to `publishAuditEvent` (`scalarAuditMap`) captured only top-level component scalars. A PATCH changing only section data produced identical `old_value`/`new_value` maps → `AuditDiff.compute` returned `null` → the SYS-048 no-op guard dropped the row. CREATE rows are persisted without a diff, which is why creation was visible while edits were not.

## Fix (SYS-053)

`sectionAuditMap` joins both snapshots of CREATE/UPDATE/RENAME events:

- BASE-configuration aspect scalars under flat dotted keys (`build.*`, `escrow.*`, `jira.*`, `baseConfiguration.versionRange`) — matching the Portal field-config paths, so the History diff reads `build.mavenVersion: 3.8 → 3.9`;
- section child collections (`vcsEntries`, `mavenArtifacts`, `fileUrlArtifacts`, `dockerImages`, `packages`, `buildToolBeans`, `requiredTools`) and per-component child collections (`artifactIds`, `securityGroups`, `teamcityProjects`, `docs`) as **content-only, deterministically ordered** snapshots — no row ids, because a PATCH REPLACE recreates child rows and id churn would turn a no-op re-save into a fake history entry (SYS-048 must keep holding).

Version-ranged rows (SCALAR_OVERRIDE / MARKER) stay out — they audit via their own field-override events (SYS-050).

## TDD

`ComponentSectionAuditTest` (ft-db integration, `dbTest`) was RED before the fix and is GREEN after: build/jira scalar PATCH, vcsEntries PATCH, requiredTools (repo-synced junction) and securityGroups (content-sorted) each write exactly one UPDATE row with a field-level diff; identical re-sends and a no-op section PATCH write nothing.

## Verification

- Full `gradlew build` green; audit-related suites (`FieldOverrideAuditTest`, `MultiValuePeopleV4Test`, `AuditLogFilterTest`, `AuditMigratedVisibilityTest`) green.
- Two-service e2e (Portal Playwright vs a locally built CRS image with this fix, full testcontainers stack): UI Build-tab edit → History shows the `build.mavenVersion` diff. Spec lands via the Portal PR (merge this one first; the spec self-skips until a CRS image with this fix is pinned).

## Docs

`requirements-common.md` SYS-053 (acceptance criteria + test traceability), `adr/005-audit-log.md` refinement.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -334,7 +334,7 @@ class ComponentManagementServiceImpl(
                 scalarAuditMap(
                     saved,
                     overrideLabels = canonicalizedLabels,
-                ),
+                ) + sectionAuditMap(saved),
         )
 
         return saved.toDetailResponse(teamcityProperties.baseUrl)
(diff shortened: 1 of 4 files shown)


pgorbachev:
@coderabbitai review

Issue #25, opened 2023-11-07
Add documentation


CamaradeRoman:
Add adoc-based specification and implement publication on wiki

Issue #24, opened 2023-10-13
Non-defined GAV should not be mapped as empty string


CamaradeRoman:
Remove temporary backward compatibility feature.

https://github.com/octopusden/octopus-components-registry-service/blob/fa682aec847bec22a8d6f50ab6a75b43b09f5aa5/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/Mappers.kt#L34

https://github.com/octopusden/octopus-components-registry-service/blob/fa682aec847bec22a8d6f50ab6a75b43b09f5aa5/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/mapper/Mappers.kt#L40

PR #360, opened 2026-06-12
fix: uniqueness gate/API to legacy-validator parity — unbricks the [1.7]/[1.8] compat stands

pgorbachev:
## Incident

The first compat run after the uniqueness gate landed (#358) failed at candidate boot: the §6.0 migration pre-pass flagged **5 violations in the production DSL** — data that passes the daily legacy validation ([2.0] Validate Production Data) and is therefore LEGAL by the de-facto contract. Two semantic regressions vs the legacy `EscrowConfigValidator` (both also latent in the v3 API checks, where they would 409 any edit of the affected components):

1. **Group matching** split BOTH sides by `[,|]` and intersected sets. Legacy `MavenArtifactMatcher` compares the row's WHOLE group string against the rival's comma-split items — CSV-vs-CSV never matches via containment, `|` is not a separator. The production model/api component families (the Defaults-inherited catch-all artifact regex vs literal placeholder artifacts under shared CSV group elements) are authored against exactly that contract → 4 false GAV collisions.
2. **Jira claims** were bucketed from raw SCALAR_OVERRIDE rows: a projectKey-only range override row carries a NULL prefix meaning *inherited*, and bucketing it as `(key, null)` falsely collided with the legitimate no-prefix owner of the same project.

A follow-up run surfaced a third regression: **2691 active diffs, all `KEY_MISSING_BASELINE at $.errorCode`** — the compat oracles byte-diff v1–v3 ERROR bodies, and every error-returning case carried the new `"errorCode": null` the production baseline doesn't have.

## Changes

- `MavenGavCollision`: group/artifact identity = UNION of legacy **#25 pattern containment** (whole-string vs comma-split items) and restored legacy **#24 exact token-pair sharing** (groups by `,`, artifacts by the legacy `[,|\s]+`, literal token equality) — the adversarial review caught that #25-parity alone would let identical-CSV duplicates through. Trim is the one documented deviation.
- New `computeEffectiveJiraPairs` (`util/JiraEffectivePairs.kt`): per-range scalar overrides layer over the BASE row (`overriddenAttribute` — not value nullness — decides overridden-ness, so null-clear works). Used by the API-side check, the migration pre-pass (row-level repository projection replaces the per-pair queries), and mirrored by the DSL collector reading loader-merged configs (loader inheritance of `componentInfo` onto projectKey-only overrides verified in `EscrowConfigurationLoader`).
- **Honest conflict messages**: rows on `group-artifact-pattern` markers are synthesized from the component-level `groupId`/`artifactId` mapping — messages now say `groupId/artifactId mapping '…'` instead of `distribution GAV '…'` (operators were hunting for a `distribution{}` block that does not exist).
- `ErrorResponse.errorCode`: `@JsonInclude(NON_NULL)` — null omitted (legacy-identical wire), set codes (v4 conflict handlers) keep emitting; pinned by `ErrorResponseSerializationTest`.
- Docs: functional-spec §2.5 updated.

## Acceptance — the real production DSL

`RealDslUniquenessAcceptanceTest` (env-gated on `REAL_CR_DSL_DIR`, skipped on CI) loads a REAL registry checkout using the checkout's OWN `gradle.properties` as the loader-parameter source and runs the pre-pass: **ZERO violations** — verified against both the exact master snapshot that bricked the original compat run and the current master.

## RED → GREEN

Commit 1 pins the legacy semantics with the production shapes reproduced 1:1 on neutral names — RED before the fix. Guard tests keep true violations caught: single group ∈ rival CSV, identical-CSV + same artifact (#24), full-GAV duplicates.

## Reviews

Sonnet correctness + Opus adversarial (both SHIP-WITH-FIXES → findings addressed: #24 restoration + tests, dead parameter, stale doc labels) + focused re-review of the fixups (SHIP). `build` + `dbTest` + `integrationTest` green.

## Operational notes

- Expected after merge: the next [1.7]/[1.8] runs go green with 0 active diffs.
- Follow-up worth wiring: export `REAL_CR_DSL_DIR` in the compat stands so the acceptance test runs where a registry checkout already exists (today it is dev-run-only).


pgorbachev:
**Latest [1.7]/[1.8] reds analysis — second root cause found and fixed here.**

The candidate now BOOTS (the uniqueness-gate false positives are gone from the DSL after the registry-side cleanup PRs). The remaining failure is different: **2691 active diffs, every single one `KEY_MISSING_BASELINE at $.errorCode`** — the compat oracles DO byte-diff v1–v3 ERROR bodies, and every error-returning case carried the new `"errorCode": null` the production baseline doesn't have.

Fixed: `@JsonInclude(NON_NULL)` on `ErrorResponse.errorCode` — null is omitted (legacy-identical wire), set codes (the v4 conflict handlers) keep emitting. Pinned by `ErrorResponseSerializationTest`. The real-DSL acceptance test is also isolated from the `ComponentsRegistryScriptRunner` product-type singleton so it runs reliably inside a shared :test JVM, and reads ALL loader parameters from the registry checkout's own gradle.properties.

This PR remains necessary even though the current DSL no longer trips the gate: the same matcher-semantics bugs sit latent in the API-side checks (any edit of similarly-shaped components would 409), and future DSL entries of the same legal shapes would brick migration again — the acceptance test proved zero violations against the original pre-cleanup master snapshot. Expected after merge: [1.7]/[1.8] go green with 0 active diffs.

(History note: the branch was rewritten to satisfy the content-validation gate — same changes, neutral names in code and commit messages.)

PR #366, opened 2026-06-25
fix: bound full /rest/api/3/components list cost at production scale (#365)

### LINKED ISSUE #365: Performance regression: GET /rest/api/3/components ~230× slower in 2.0.88-3984 vs 2.0.87
## Summary

The full component list endpoint `GET /rest/api/3/components` is **~230× slower** in `2.0.88-3984` than in `2.0.87`, for a byte-for-byte identical response. This is a server-side processing regression — not network, payload size, or pod resources.

## Measurements

Measured from **inside the CRS pod** against `localhost:8080` (no network, no auth, anonymous — the same way the portal's validation sweep calls it), while the service was idle:

| Environment | CRS version | `time_total` | response size |
|---|---|---|---|
| prod | `2.0.87` | **~0.015 s** (3 runs: 0.020 / 0.013 / 0.012) | 1 327 246 bytes |
| test/QA | `2.0.88-3984` | **~3.5 s** (3 runs: 3.79 / 3.42 / 3.31) | 1 327 246 bytes |

Same endpoint, **identical payload size** (1 327 246 bytes, ~988 components), idle service, localhost → ~230× slowdown is intrinsic to request handling in `2.0.88-3984`.

## Repro

From inside the running CRS pod:

```
curl -s -o /dev/null -w 'http=%{http_code} time=%{time_total}s size=%{size_download}\n' \
  http://localhost:8080/rest/api/3/components
```

Run against `2.0.87` (≈15 ms) and `2.0.88-3984` (≈3.5 s).

For contrast, the **paged** endpoint is fast on the same `2.0.88-3984` instance:

```
GET /rest/api/4/components?page=0&size=20&archived=false   → ~0.13 s, ~11 KB
```

So the regression is specific to building/serializing the **full unpaged listing**, not component access in general.

## Impact

Under any concurrent background load on the instance (startup DSL→DB migration, TC resync), the already-slow full-list latency balloons past 30 s, which causes:

- Downstream consumers that fetch the full list to time out (e.g. the portal's scheduled validation sweep aborts on the component-list fetch at its 30 s per-request timeout).
- UI list views to hang while the service is busy.

Both disappear on `2.0.87`, where the list returns in ~15 ms.

## Suggested investigation

- Diff request handling for `/rest/api/3/components` between `2.0.87` and `2.0.88-3984` (serialization, per-element work, N+1 queries, eager variant/aspect loading, added per-component computation).
- Why an idle, localhost, identical-size response costs ~3.5 s of CPU — likely per-element overhead introduced in the 2.0.88 line.

## Severity

High for environments running the 2.0.88 line: it degrades full-list consumers and UI under normal background activity. Prod (`2.0.87`) is unaffected.

pgorbachev:
## Summary

Fixes #365 — the full unpaged listing `GET /rest/api/3/components` was ~230× slower on the DB-backed resolver (`2.0.88`) than the git-backed one (`2.0.87`).

**Root cause (architecture shift, not a one-method bug):** `2.0.87` serves this list from a startup-built in-memory graph (O(1), ~15 ms). `2.0.88` routes to `DatabaseComponentRegistryResolver.getComponents()`, which **rebuilds the entire merged domain model for every component on each request**. Two compounding costs scaled with production size:

1. **DB round-trips.** `@BatchSize` was `100`, *below* the ~988 prod component count, so every lazy role took `ceil(owners/100)` `IN`-selects ≈ **~300 round-trips/request** — dominating latency against a remote DB and ballooning past 30 s under concurrent migration/sync load. (`@EntityGraph`/join is blocked by `MultipleBagFetchException`, so batch sizing is the lever.)
2. **CPU/reflection.** `EntityMappers.setField` called `getDeclaredField` uncached ≈ **~50k reflective lookups/request**.

## Changes (surgical scope only)

- **`BATCH_FETCH_SIZE` 100 → 1000** (above the prod component count). Audited that every role walked by `getComponents()`/`toEscrowModule` references this constant:

  | Role group | `@BatchSize` source |
  |---|---|
  | `ComponentEntity` `@OneToMany` (configurations, artifactIds, securityGroups, teamcityProjects, docLinks, releaseManagers, securityChampions, labelJunctions) | field-level `@BatchSize(BATCH_FETCH_SIZE)` |
  | `ComponentConfigurationEntity` `@OneToMany` (vcsEntries, mavenArtifacts, fileUrlArtifacts, dockerImages, packages, requiredToolJunctions, buildToolBeans) | field-level `@BatchSize(BATCH_FETCH_SIZE)` |
  | to-one targets `ComponentEntity` (parentComponent), `ComponentGroupEntity`, `ToolEntity` | class-level `@BatchSize(BATCH_FETCH_SIZE)` |

  `hibernate.default_batch_fetch_size: 50` is only the fallback for un-annotated roles; field/class-level `@BatchSize` takes precedence for all walked roles.
- **Memoized reflective field lookups** — `escrowModuleConfigField` builds the `EscrowModuleConfig` field map once at class load (Groovy synthetic fields filtered); `setField` and the resolver's `distribution` write reuse it (the latter `requireNotNull`, fail-loud).

## Tests (test-first)

- **`GetComponentsListQueryCountTest`** (`dbTest`, blocking gate): batch-boundary guard at 150 vs 450 components asserts **equal, size-independent** statement counts. **RED pre-fix: 150→39, 450→108. GREEN post-fix: 150→16, 450→16.**
- **`GetComponentsListPerformanceTest`** (`@Tag("performance")`, **non-gating** `perfTest` task — excluded from `test`/`build`/`dbTest`): ~1k components → ~29 statements, **~166 ms median** (local Testcontainers; a generous 2 s SLA ceiling guards against gross CPU regressions, calibrate per env).
- **`EntityMappersSetFieldCacheTest`**: field discovered once per name; unknown/synthetic names → null.

## Notes

- Surgical fix does **not** reach `2.0.87`'s ~15 ms (still rebuilds per request), but removes the round-trip storm. Full parity via a result cache is deferred — see `docs/registry/tech-debt/014-get-components-result-cache.md` (a naive `count + max(updated_at)` probe is unsafe because child/junction writes don't reliably bump parent `updated_at`).
- Local pre-commit `./gradlew build` green; `dbTest` query-count guards green; `perfTest` green.

### DIFF
--- a/.github/workflows/perf-test.yml
+++ b/.github/workflows/perf-test.yml
@@ -1,0 +1,37 @@
+# INTERNAL / NON-GATING — manual + weekly schedule only; never on pull_request, never part of the
+# merge gate. Runs the @Tag("performance") SLA guard (GetComponentsListPerformanceTest, GH #365),
+# which is excluded from `test`/`build`/`dbTest` because its wall-clock ceiling is timing-sensitive.
+# This workflow makes the otherwise-never-executed `perfTest` a real, observable regression signal
+# without ever blocking a merge.
+name: Performance SLA (non-gating)
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Mondays 05:00 UTC. Reports a perf regression; does not gate any branch.
+    - cron: '0 5 * * 1'
+
+jobs:
+  perf-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '21'
+
+      - name: Run performance SLA guard
+        run: ./gradlew :components-registry-service-server:perfTest --no-daemon
+
+      - name: Publish perf test report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: perf-test-report
+          path: |
+            components-registry-service-server/build/reports/tests/perfTest/**
+            components-registry-service-server/build/test-results/perfTest/**
+          if-no-files-found: warn
--- a/build.gradle
+++ b/build.gradle
@@ -230,7 +230,11 @@ configure(subprojects) {
     // (hence from `check`/`build`) and run in `dbTest` — CI [1.2] / GH qualityCoverage.
     test {
         useJUnitPlatform {
-            excludeTags 'integration'
+            // `performance` (GH #365 perf SLA guard) is heavy (Testcontainers + ~1k-component
+            // fixture) and timing-sensitive — excluded from the fast gate AND from `dbTest`; it
+            // runs only under the dedicated `perfTest` task so a flaky timing ceiling never
+            // blocks normal CI.
+            excludeTags 'integration', 'performance'
         }
     }
 
@@ -247,6 +251,23 @@ configure(subprojects) {
             shouldRunAfter tasks.named('test')
             useJUnitPlatform {
                 includeTags 'integration'
+                excludeTags 'performance'
+            }
+        }
+    }
+
+    // Non-gating performance SLA guard (GH #365). Separate task so a timing-sensitive ceiling
+    // never blocks `check`/`build` or the `dbTest` correctness gate. Run explicitly:
+    //   ./gradlew :components-registry-service-server:perfTest
+    if (project.name == 'components-registry-service-server') {
+        tasks.register('perfTest', Test) {
+            description = 'Runs @Tag("performance") SLA guards (Testcontainers/Spring); NOT in check/dbTest.'
+            group = 'verification'
+            testClassesDirs = sourceSets.test.output.classesDirs
+            classpath = sourceSets.test.runtimeClasspath
+            shouldRunAfter tasks.named('test')
+            useJUnitPlatform {
+                includeTags 'performance'
             }
         }
     }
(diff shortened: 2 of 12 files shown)


pgorbachev:
Addressed all three review points (force-pushed `242ef5b3`):

**[P2] `perfTest` now runs automatically** — added `.github/workflows/perf-test.yml`: a separate **non-gating** workflow (`workflow_dispatch` + weekly `schedule`, Mondays 05:00 UTC) that runs `./gradlew :components-registry-service-server:perfTest` and uploads the report. It is not referenced by `gate/merge`, so the non-gating semantics are preserved while the test becomes a real observation signal.

**[P2] Query-count guard now enforces production-safe batch size** — `LARGE` raised `450 → 500`, so config-owned roles have **exactly 1000 owners**, and the invariant is now `LARGE * 2 <= BATCH_FETCH_SIZE` (= 1000). Any `BATCH_FETCH_SIZE` below 1000 (i.e. below the ~988 production component count) pushes the LARGE config-owned roles into a second `IN` batch at LARGE but not SMALL → the equality assertion fails. So the test now guards the design decision behaviorally, not just the constant. Re-verified: SMALL(150)=16, LARGE(500)=16 (GREEN).

**[P3] Stale doc fixed** — `DatabaseComponentRegistryResolverQueryCountTest` no longer hardcodes `@BatchSize(100)`; it references `BATCH_FETCH_SIZE` and points to `GetComponentsListQueryCountTest` for the above-threshold boundary behavior.

Issue #357, opened 2026-06-12
v4 write path does not validate the artifactIds ownership mapping (legacy #24/#25 subject)


pgorbachev:
## Context

The legacy pre-publish validator's rules #24/#25 (`EscrowConfigValidator.validateGroupIdAndVersionIdIntersections` / `validateVersionConflicts`) enforced that no two components claim intersecting `groupId`/`artifactId` ownership-mapping patterns in intersecting version ranges — this mapping drives "which component owns this artifact" resolution.

In v3:
- The distribution-GAV cross-component check (`validateMavenArtifactCollisions`, full `(group, artifact, extension, classifier)` identity since the 2026-06-12 fix) covers `distribution { GAV }` rows **and** the per-range override mapping (MIG-047 `GROUP_ARTIFACT_PATTERN` rows live in the same table).
- BUT the **base-row ownership mapping** (`component_artifact_ids`, `ComponentArtifactIdEntity`) is written by both the import and the v4 API (`ComponentManagementServiceImpl` ~line 993) **without any cross-component intersection validation** — the repository is not referenced by `validateCrossComponentIntegrity` at all.

## Impact

Two components can claim intersecting artifact-ownership patterns via the v4 API; artifact→component resolution (escrow/jira paths) then becomes ambiguous — the exact failure mode legacy #24/#25 existed to prevent.

## Proposed fix

Add a cross-component intersection check over `component_artifact_ids` (same pattern semantics as `MavenGavCollision.patternsOverlap` + version-range intersection) to `validateCrossComponentIntegrity`, and the matching invariant to the §6.0 migration uniqueness pre-pass. Scan prod DSL for pre-existing violations first.

## References

- `.github/audit/VALIDATION-PARITY-2026-06-03.md` — Addendum 2026-06-12.
- `docs/registry/functional-spec.md` §2.5.

pgorbachev:
Implemented and merged.

- CRS: #373 → `v3` (squash `2d6f6208`) — explicit artifact-ID ownership modes (EXPLICIT / ALL_EXCEPT_CLAIMED / ALL) on the new `component_artifact_mappings` tables, with mode-aware cross-component uniqueness enforced on the v4 write path (the subject of this issue), a strict DSL→mode migration classifier, and behavioral v1–v3 compatibility.
- Portal: #127 → `develop` (squash `8fa51b65`) — the artifact-ID ownership editor (mode radio, per-range overrides, conflict UI).

[...]

PR #373, opened 2026-06-26
feat(#357): explicit artifact-ID ownership model (modes) + deterministic uniqueness

pgorbachev:
## What & why

Issue #357 asked to restore the legacy cross-component artifact-ownership uniqueness checks (#24/#25) that v3 had dropped. While scoping it we found the underlying field — the base groupId/artifactId mapping, stored as an opaque regex `component_artifact_ids.artifact_id_pattern` — was the real blocker: you cannot decide ownership uniqueness precisely while ownership is an opaque pattern, the inherited default surfaced as a raw `[\w-\.]+` in the UI, and the per-range override of the pattern was import-only (not editable via v4). So this models ownership **explicitly** and restores the checks deterministically.

## Model

A component owns a **list of mappings**; each mapping = comma-separated **group list** + an ownership **mode** + a **version range** (+ literal **tokens** for EXPLICIT):

| Mode | Owns | Legacy origin |
|------|------|---------------|
| `EXPLICIT` | exactly the listed literal artifacts under its group(s) | literal / `,`/`\|` enumeration |
| `ALL_EXCEPT_CLAIMED` | any artifact under its group NOT explicitly claimed by another component (yields) — single-group | the `(?!X)` negative-lookahead |
| `ALL` | every artifact under its group(s), unconditionally | inherited catch-all / no explicit artifactId — the `main` default |

New tables `component_artifact_mappings` + `component_artifact_mapping_tokens` (drop `component_artifact_ids`; retire the `GROUP_ARTIFACT_PATTERN` ownership marker). A per-range override **REPLACES** the base for its range (most-specific wins); `sort_order=0` = primary.

## Uniqueness (restores #24/#25)

Decided deterministically from stored modes (no regex probing): `EXPLICIT×EXPLICIT` conflicts iff tokens intersect; `EXPLICIT×ALL_EXCEPT_CLAIMED` never; `ALL_EXCEPT_CLAIMED×ALL_EXCEPT_CLAIMED` conflicts; `ALL×anything` conflicts.

**Enforcement boundary (decision):** runs on the **v4 write path only** — create (unconditional) and update when `artifactIds` is present. A field-override cannot change ownership and never re-triggers it (this was a deliberate fix — re-running the matrix from the shared field-override path 409'd on pre-existing data). It is **not** wired into the §6.0 migration pre-pass: production is overlap-free by construction under the legacy single-match resolver, and migration enforces correctness via strict classification instead. The self side uses **effective-per-range** claims so an override that narrows tokens does not false-409 a rival owning the freed token in that range.

## Migration / v1–v3 compat

Strict ordered classifier (no escape hatch): `(?!`→ALL_EXCEPT; exact catch-all set→ALL; literal/CSV/pipe enumeration→EXPLICIT; any other regex **hard-fails**. v1–v3 wire renders the **primary** mapping per range; reverse `find-by-artifact` flattens **all** mappings; EXPLICIT tokens are dot-escaped at the legacy regex boundary and matched by exact equality in the DB resolver. Multi-mapping components are deterministically lossy forward, complete reverse — all imported data is single-mapping so behavior is unchanged.

## v4 API / DTO

`ArtifactIdRequest`/`Response` carry `mode` + `artifactTokens`; the response adds read-only server-computed `legacyArtifactIdPattern`. PATCH `artifactIds` is a full replacement; `sort_order` is server-derived from list position. Write-time invariants (400): group/token allowlist `[A-Za-z0-9_.-]+`, EXPLICIT≥1 token / ALL{,_EXCEPT} tokenless, ALL_EXCEPT single-group, range alignment to an existing config range, disjoint override ranges, intra-component group-token disjointness.

## Docs
functional-spec §2.5, SYS-056 (`requirements-common.md`), ADR-017, validation-parity addendum.

## Verification
- Full `gradlew build` + full `:dbTest` integration suite green.
- Prod-DSL gate `RealDslUniquenessAcceptanceTest` (against the real production registry checkout): **0 unclassifiable patterns, 0 §6.0 uniqueness violations**.
- Independent reviewer pass; the surfaced P1/P2 findings (misleading migration comment, mid-loop `sort_order` query, effective-per-range replacement in the uniqueness check) are fixed in this branch.

## Follow-up
- Portal PR B (ownership editor: mode radio, token chips, per-range override, conflict banner) — separate, base `develop`, after this merges.
- Rival-side effective-per-range range-restriction (needs rivals' config ranges) — sound today (only over-reports in a doubly-rare symmetric override-collision shape prod doesn't exhibit); tracked.


pgorbachev:
Addressed review findings (pushed in `66e39b95`, merged latest `v3`):

- **Rival-side false-409 (override REPLACES base)** — a base (`ALL_VERSIONS`) ownership claim is now skipped against any range its *own* component overrides, applied symmetrically to the edited component **and** every rival via a per-component shadow-range map derived from the already-loaded mappings (no extra query). Sound: a base is skipped only where it is provably replaced; a strict-subset override range only over-reports, never misses a real conflict. Regression-guarded in `OwnershipCollisionMatrixTest` (shadow-skip + a does-not-mask-a-real-conflict case).
- **DSL/code export + UI preview of `ALL_EXCEPT_CLAIMED`** — now renders the anchored negative-lookahead over the in-force EXPLICIT siblings (`(?!(?:s1|s2)$)[\\w-\\.]+`) instead of the plain catch-all, so exported DSL no longer reintroduces the overlap the mode encodes away. New `ArtifactOwnershipRendering.renderExportPattern`; the v3 API **wire** form is unchanged (catch-all; resolver yields via specificity).
- **`renderFull`/`renderResolved` ownership** — the full code view now emits a range block for an ownership-only per-range override (previously dropped after ownership left the `GROUP_ARTIFACT_PATTERN` marker), and the resolved view shows the effective (override-replaces-base) ownership for the version. Renderer stays pure; the service passes a precomputed sibling-aware export-pattern map (gated on `ALL_EXCEPT_CLAIMED` presence — common path issues no extra query).
- **`legacyArtifactIdPattern` (detail response)** — enriched server-side with the sibling-aware lookahead for `ALL_EXCEPT_CLAIMED` (the wire mapper only knows the catch-all; the preview needs cross-component siblings).

Full build + full `dbTest` + prod-DSL gate (`RealDslUniquenessAcceptanceTest`) green. New tests: `ArtifactOwnershipRenderingTest`, extended `OwnershipCollisionMatrixTest` + `ComponentCodeRendererTest`. SYS code renumbered to **SYS-058** (v3's #371 took 056/057).

pgorbachev:
Addressed the follow-up review (pushed `7318c4d2`):

- **Resolved v1–v3 DTO ignored per-range ownership** — `getResolvedComponentDefinition(comp, v)` selected ownership by `base.versionRange` (ALL_VERSIONS), returning base groupId/artifactId at a version an override owns (disagreeing with `renderResolved`). `buildEscrowModuleConfig` now takes an `ownershipRange`; the resolved path passes the narrowest ownership-override range containing the version (else base). Per-range enumeration path unchanged (defaulted).
- **Deterministic ownership mapping order** — base and per-range buckets both start at `sortOrder=0`; v4 detail + audit now use a shared `ARTIFACT_MAPPING_ORDER` comparator (base → versionRange → sortOrder → id) instead of `sortBy(sortOrder)`.
- **ALL_EXCEPT sibling preview** — excludes this component's own EXPLICIT rows (they replace, not narrow, the base mapping); only other components' EXPLICIT claims feed the lookahead.

**Migration pre-pass ownership (raised again):** kept v4-write-path-only by design, and clarified in-code why it is *not* a gap — migrated DSL is legacy-valid, and an ownership collision is exactly what the legacy single-match resolver throws on, so it cannot exist in migrated data; only a new v4 write can introduce one (guarded by the matrix). Strict artifactId→mode classification (hard-fail, no escape hatch) covers migration correctness. Happy to wire the matrix into the §6.0 pre-pass too if defense-in-depth is preferred — it would require making the ft-db boot fixtures ownership-clean.

Full build + full `dbTest` + prod-DSL gate green.

PR #382, opened 2026-06-29
fix(registry): single-delta supported-versions PUT — no flush violation on coarse re-PUT (ADR-018) — PR-3c

pgorbachev:
## Summary

Follow-up to PR-3b (#381). A re-review found a remaining **unique-index risk** the first delta fix missed: a **coarse PUT over already-split coverage** could trip `uq_component_configurations_one_range_presence` (500).

### The bug
Component has override `[2,3)` and supported rows `[1,2)`, `[2,3)`, `[3,10)` (from a prior auto-split). `PUT {ranges:["[1,10)"]}`: the delta step removed the three split rows and added `[1,10)`, then the **separate** `autoSplitCoverage` removed `[1,10)` and re-added `[1,2)`, `[2,3)`, `[3,10)` — re-adding the **identical** `(component, range)` rows that were just orphan-removed, in the same flush. Hibernate orders INSERTs before the orphan-removal DELETEs → duplicate-key violation.

### The fix
Compute the **final post-split coverage up front** (new pure `coverageAfterOverrideSplit`: split the requested ranges at every single-segment override edge; reject a composite covering range intersected by a simple override), then apply **one delta** against the current rows — identical rows are **kept**, never removed-then-re-added. A coarse re-PUT is now a true no-op. `setSupportedVersions` no longer calls `autoSplitCoverage` (the field-override write paths still do).

## Testing
- `SupportedVersionsApiTest`: new coarse-re-PUT-over-split regression (200, re-aligned rows, no 500) + existing idempotent/overlap/sentinel/re-align cases.
- Full unit + `:dbTest` green.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -1001,6 +1001,40 @@ class ComponentManagementServiceImpl(
      * old ones back, so repeated range edits fragment `RANGE_PRESENCE` over time. Correctness holds
      * (union + per-row constant values); a compaction pass is a follow-up.
      */
+    /**
+     * Pure auto-split (ADR-018 (b)) of a requested supported set: split each requested range at every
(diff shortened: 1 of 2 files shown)


pgorbachev:
Superseded by the version-model redesign (enumeration decoupled from coverage): write-time auto-split is removed, so this coarse-PUT delta fix is moot. Folded into the single consolidated redesign PR. Closing without merge.

PR #403, opened 2026-07-04
fix(v4): restore multi-system membership via component_systems junction

pgorbachev:
## What & why

A component may be classified under **several system codes at once** (the DSL `system = "X,Y"` CSV, and ≥174 such components exist in one production installation). PR #301 had collapsed the M:N `component_systems` junction to a single scalar `components.system_code`, so DB-mode import kept only the **first** DSL system and silently dropped the rest. A component was then missing from any report or `?system=` filter for its *other* systems.

This surfaced as two failing FT tests in the downstream releng system-release / IPS reports. The affected system codes are not exercised by production traffic, so the request-replay compat baseline CRS gates merges on never touched this path — the regression went unnoticed.

This PR restores the many-to-many model, coherently across every layer.

## Changes

- **Schema** — reinstate the `component_systems` junction (drop the scalar column, its index, and FK). DB is recreated on deploy, so `V1__schema.sql` is edited in place (no incremental migration, per project policy).
- **Entity/repo** — `ComponentSystemEntity` + `ComponentSystemRepository` (mirrors the labels junction); `ComponentEntity.systemJunctions`.
- **Import** — `linkSystems` persists **all** CSV codes (deduped); no first-only drop, no WARN.
- **Read** — v1/v2/v3 (`Component.system: Set<String>`) and v4 (`systems: Set<String>`) return the full set; `?system=A,B` filters via a junction JOIN + `distinct` (OR-semantic).
- **Write** — v4 create/update accept `systems`; each code validated against the config allowlist **or** the master `systems` dictionary, canonicalised (case-insensitive → canonical casing); field-config-hidden strips the write; audit records the persisted set.
- **Ordering** — membership is order-agnostic; every consumer uses `contains(...)`, and git-vs-DB parity is set-equality. No `sort_order` column and no ordering guarantee is imposed (matches the pre-#301 behaviour and the investigation's ordering-semantics analysis).
- **Docs** — ERD, schema-spec, requirements-common, and the OpenAPI v4 spec updated.

## Tests (RED → GREEN)

- `MultiSystemMembershipTest` (new) — v4 round-trip guard: a component created with two systems is returned by a filter for **either** system, and detail/summary echo both. Committed first as RED (fails against the scalar model), green after the fix.
- `StrictContractTest` — v4 `systems` create (single/multi/empty/reject/casing), PATCH (replace/no-op/casing), field-config-hidden audit ghost-write guard.
- Multi-system git-vs-DB parity is covered by the existing `TESTONE` fixture (`system = "CLASSIC,ALFA"`) exercised by both the git-source and DB-source controller tests, which now both assert the full `{ALFA, CLASSIC}` set.
- Full `build` + full server `dbTest` suite green.

## Notes / follow-ups

- The **field-config key stays `component.system`** (opaque identifier); the API field is `systems`. Aligning the key to `component.systems` is deferred to the companion PRs so admin config-as-code stays in sync.
- **Companion PRs** (standard split): Portal (`develop`) — multi-select UI + type regen after `crs.version` bump; `service-config` — field-config key rename. Portal issue linked below.


pgorbachev:
Portal companion issue (UI multi-select + type regen; and the `service-config` field-config key rename `component.system` → `component.systems`): octopusden/octopus-components-management-portal#158

Merge order: this CRS PR → `v3` first, then bump `crs.version` in the Portal companion.

pgorbachev:
Addressed Copilot review:
- **MultiSystemMembershipTest KDoc grammar** — fixed (4eb89e6d).
- **"blank line after package" on ComponentSystemEntity/ComponentSystemRepository** — intentionally left: both mirror the existing `ComponentLabelEntity`/`ComponentLabelRepository` style (no blank line after `package`), and the ktlint gate (`quality/static`) is green. Adding a blank line would make them inconsistent with the established junction-file convention.

Issue #402, opened 2026-07-04
Version-render endpoint for live editor preview (unsaved formats + per-range overrides)


pgorbachev:
## Context

The portal's Jira-tab **Version Preview** shows how each version format renders (release / rc / minor / line / build / hotfix). For standard build systems the portal computes this client-side, but that's a naive template string-swap — it can't reproduce build systems whose scheme is library-driven (Whiskey: zero-padding like `03.62.30.19-4`, custom `$…Cards` variables, computed segments).

As a first step the portal now calls the existing legacy endpoint for Whiskey:
`GET /rest/api/2/components/{component}/versions/{version}/detailed-version` → `DetailedComponentVersion`.
That is correct but renders the **saved** configuration only — it can't reflect **unsaved** format edits happening live in the editor, and it doesn't let the editor preview per-range overrides for a chosen version.

(Portal side already shipped: octopusden/octopus-components-management-portal#156.)

## Goal

A render endpoint that lets the editor show a **live, version-accurate** preview from the **unsaved** configuration, including **per-range overrides**.

## Proposal — Variant A (server resolves the range)

Add `POST /rest/api/4/versions/preview`.

The client sends the **current unsaved** Jira configuration — the base formats **and** the per-range override rows — plus the build system, version names, and an input version. The server:

1. **Resolves which range the input version falls into** and applies the matching override (or base) — range resolution stays server-side, the single source of truth (the client must not duplicate it).
2. Renders the effective formats with the real `JiraComponentVersionFormatter` (`org.octopusden.releng.versions`), honouring the build system and `VersionNames`.
3. Returns the version ladder.

### Request (sketch)

```jsonc
{
  "version": "1.0.50",                 // input version — drives range selection
  "buildSystem": "WHISKEY",
  "versionNames": { "serviceBranch": "serviceCBranch", "service": "serviceC", "minor": "minorC" },
  "technical": false,
  "base": {
    "versionPrefix": "pgw",
    "versionFormat": "$versionPrefix-$baseVersionFormat",
    "lineVersionFormat": "$major.$minor",
    "minorVersionFormat": "",          // empty = mirrors line
    "releaseVersionFormat": "$major.$minor.$service-$fix",
    "buildVersionFormat": "",          // empty = mirrors release
    "hotfixVersionFormat": "$major.$minor.$service-$fix-$build"
  },
  "overrides": [
[...]

PR #404, opened 2026-07-04
feat(versions): live version-format preview endpoint (POST /rest/api/4/versions/preview)

pgorbachev:
## What

Adds `POST /rest/api/4/versions/preview` — a **stateless** endpoint that renders a `DetailedComponentVersion` from ad-hoc Jira version formats (a `base` block + per-range `overrides`) and an input version, with **no persistence and no component lookup**. It lets the Portal component editor show a **live, version-accurate** preview of the *unsaved* config, instead of the saved-config `detailed-version` call (CRS issue #402).

## Parity by construction

Preview output equals `GET .../detailed-version` for the same effective config + version, because it reuses the persisted render path:
1. Resolve the range server-side (first override whose range contains the version wins; base otherwise).
2. Materialise the effective `ComponentVersionFormat` — `line`/`build` mirror `minor`/`release`, and `minor`/`release` default to `$major`/`$major.$minor` when blank (matching `EntityMappers.buildJiraComponent`).
3. Canonicalise the version with the same `JiraComponentVersionFormatter.normalizeVersion(..., strict=false, hotfixEnabled)` the resolver's `getJiraComponentVersion` uses, then drive the existing `Mapper<JiraComponentVersion, DetailedComponentVersion>`.

## Notes

- **Not DB-gated** — only needs the always-present `VersionNames`/formatter beans, so it works in git and DB modes.
- **Auth**: authenticated-only via the existing `rest/api/4` catch-all; no finer permission since no persisted data is touched.
- **`buildSystem` deliberately absent** — padding / custom-prefix wrapping is driven purely by the format templates + `VersionNames`.
- **Hotfix** eligibility is caller-supplied `hotfixEnabled` (VCS-branch-derived in the persisted path), not inferred from a hotfix format's presence.
- **4xx**: blank/non-numeric version, malformed range, or prefix-without-format → `400`; version matching no configured format → `404`.
- Field naming mirrors the v4 write contract (`minorVersionFormat` ↔ the resolver's `majorVersionFormat`).

## Tests / gates

- Unit (`VersionPreviewServiceImplTest`): base render, override-range selection, custom prefix (`.version` vs wrapped `.jiraVersion`), template-driven padding, hotfix gating, unmatched→404, invalid/malformed→400.
- Integration (`VersionsControllerV4Test`): 200 authenticated, 401 anonymous, 400 invalid, **and a parity guard** asserting preview == `detailed-version` for a real seeded component (custom prefix + hotfix).
- OpenAPI v4 spec regenerated (`OpenApiV4SpecTest` green); `qualityStatic` green.
- New requirement **SYS-059** in `docs/registry/requirements-common.md`.

This must land in **v3** first so the Portal can regenerate its vendored spec against it (the portal switch is a follow-up PR).

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/config/OpenApiV4Config.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/config/OpenApiV4Config.kt
@@ -39,6 +39,8 @@ class OpenApiV4Config {
                 "/rest/api/4/info",
                 // MigrationStatusControllerV4 — anonymous migration-activity probe (SYS-055).
                 "/rest/api/4/migration-status",
+                // VersionsControllerV4 — stateless version-format preview (SYS-059).
+                "/rest/api/4/versions/**",
                 // AuthController at /auth/me — outside /rest/api/4.
                 "/auth/**",
             )
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/VersionsControllerV4.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/VersionsControllerV4.kt
@@ -1,0 +1,59 @@
+package org.octopusden.octopus.components.registry.server.controller
+
+import io.swagger.v3.oas.annotations.Operation
+import org.octopusden.octopus.components.registry.core.dto.DetailedComponentVersion
+import org.octopusden.octopus.components.registry.server.dto.v4.VersionPreviewRequest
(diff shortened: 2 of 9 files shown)


pgorbachev:
Addressed Copilot's review in f60ef1d6:
1. `VersionsControllerV4` — per-request log moved from INFO → DEBUG (the editor calls this on every debounced keystroke).
2. `hotfixVersionFormat` @Schema — corrected: a hotfix coordinate is gated on `hotfixEnabled`, not on the format's presence.
3. `hotfixEnabled` @Schema — corrected: when true and no format resolves, a degenerate empty hotfix coordinate is rendered for detailed-version parity.
`openapi/v4.json` regenerated; OpenApiV4SpecTest + qualityStatic green.

Issue #387, opened 2026-07-01
Migration: reject per-range distribution.explicit / distribution.external / securityGroups.read 


pgorbachev:
## Problem

`distribution.explicit` and `distribution.external` in the DSL are **per-component** in the
schema-v2 / ADR-018 model, but the DSL grammar still accepts `distribution { explicit = ... }`
and `distribution { external = ... }` **inside a bounded-range block**. When the per-range value
differs from the top-level / base value, the import path silently drops the per-range value and
resolve returns the base — a data-loss class of bug.

Discovered while running `octopus-dms-service` FT compat against CRS `2.0.88-4086`. The
`ee-component-with-version-ranges` test fixture (test-common, PR
octopusden/octopus-dms-service#87 by @rodemark) exercises exactly this shape:

```groovy
"ee-component-with-version-ranges" {
    distribution { external=true,  explicit=false, docker="test/test-component-vr" }        // top-level
    "[03.53,03.54)" { distribution { external=true,  explicit=false, docker="test/test-component-vr" } }
    "[03.54,03.55)" { distribution { external=true,  explicit=true,  docker="test/test-component-vr" } }  // <-- explicit=true only here
    "[03.55,03.56)" { distribution { external=false, explicit=true,  docker="test/test-component-vr" } }
}
```

- Test `testRegisterComponentVersionArtifactUsesVersionRangeDistribution[2] 03.54.30.64, null`
  expects DMS to see `distribution.explicit=true` (from range `[03.54,03.55)`); it gets `false`.
- CRS `Get Component: ee-component-with-version-ranges:03.54.30.64` returns base `explicit=false`.

## Where the loss happens

`EntityMappers.buildEscrowModuleConfig` (`EntityMappers.kt:699-708`):
```kotlin
val distribution =
    buildDistribution(
        explicit = component.distributionExplicit,   // per-component scalar column
        external = component.distributionExternal,   // per-component scalar column
        mavenArtifacts = mavenArtifacts,             // routed via pickMarkerChildren -> per-range
        fileUrlArtifacts = fileUrlArtifacts,         // routed via pickMarkerChildren -> per-range
        dockerImages = dockerImages,                 // routed via pickMarkerChildren -> per-range
        packages = packages,                         // routed via pickMarkerChildren -> per-range
        securityGroups = component.securityGroups.toList(),   // per-component
    )
```

Write side (`ImportServiceImpl.kt:1329-1333`):
```kotlin
[...]

PR #406, opened 2026-07-04
fix(dist): fail-loud guard on malformed distribution.GAV coordinates (TD-011)

### LINKED ISSUE #349: TD-011: groupId-only Maven GAV round-trip dropped by schema-v2 (SYS-030)
Promotes tech-debt doc `docs/registry/tech-debt/011-gav-group-id-only-support.md` to a tracked issue.

## Problem

V1 Groovy DSL supports a `distribution.GAV` value that is a **bare groupId** (no artifactId), e.g. `"org.example.teamcity.ee"`. Schema-v2's import + resolver pipeline **silently drops** these entries → value-level divergence on distribution endpoints (`/components/{name}` v2, `/components/{name}/distribution`).

## Verified live on v3 HEAD

- `DistributionMavenArtifactEntity.artifact_pattern` is `TEXT **NOT NULL**` (line 27) — no representation for "no artifactId".
- `GavParsing.parseMavenGavEntry()` returns `null` for <2 `:`-segments; `ImportServiceImpl.attachMavenArtifacts` does `parseMavenGavEntry(entry) ?: continue` → groupId-only entries are dropped at import.

## Cutover relevance / why compat may be silent

Divergence is visible only for components whose DSL uses a bare groupId in `distribution.GAV`. If no such component is in the smoke set or the trace-replay window, compat shows nothing. **First step: confirm whether any current DSL component uses a bare-groupId `distribution.GAV`** — if none, downgrade; if yes, it's a real cutover regression and trace-replay should surface it as STRUCTURAL/VALUE_DIFF.

## Fix (per doc — ~10 files, single PR)

Nullable `artifact_pattern` path: drop `NOT NULL`, `MavenCoords.artifactId: String?`, parser returns `MavenCoords(groupId, null, …)` for 1-segment input, import writes `artifactPattern = null`, read/compose handles null → emits pure groupId, V4 DTO + write-validation updates, round-trip + V4-write tests. Full spec in the doc.

## Acceptance

- Fixture/compat test pins V1 behaviour for a known groupId-only component.
- v2 resolver returns the same `ComponentArtifactConfiguration` shape as V1 byte-for-byte.
- Round-trip: bare groupId → import → `artifact_pattern = NULL` → resolver → same bare groupId.

Source: `docs/registry/tech-debt/011-...md` (SYS-030).

pgorbachev:
Refs #349 (TD-011). Closes the silent-drop of unparseable Maven `distribution.GAV` coordinates, symmetrically on the import path and the v4 write path.

## Problem
`ImportServiceImpl.attachMavenArtifacts` did `parseMavenGavEntry(entry) ?: continue` — a `distribution.GAV` value that isn't `groupId:artifactId[:ext[:classifier]]` (a groupId-only value with no `:`, or a blank segment) was **silently dropped** at import → value-level divergence on the distribution endpoints, invisible to the compat baseline. The v4 write path was the symmetric gap: `replaceMavenArtifacts` copied `groupPattern`/`artifactPattern` verbatim, so a blank coordinate was accepted (201).

## Change
- **Import (fail-loud):** new `validateDistributionCoordinates(componentKey, configs)` mirroring the #387 `PerComponentDistributionGuard` — a non-URL `distribution.GAV` entry that fails to parse now fails that component's import loudly (component + raw entry + reason), instead of vanishing. Invoked in `importModule` beside the #387 guard, so it runs in the same per-component try/catch (fails just that component, not the batch — same blast radius as #387).
- **v4 write (symmetric):** `validateMavenArtifactCoordinate` rejects blank `groupPattern`/`artifactPattern` on create/PATCH/marker overrides (400), via `replaceMavenArtifacts` (the single funnel).
- **Scope:** DEB/RPM are plain package filenames (not `:`-coordinates) and can't silent-drop, so they're intentionally out of scope; docker accepts a no-colon name as a valid image.

## Decision (fail-loud vs TD-011 "support")
TD-011's doc proposes *supporting* a groupId-only GAV (nullable artifactId, round-tripped). This PR takes the **fail-loud** resolution instead (per request): convert silent data-loss into a clear, actionable failure. No in-repo DSL uses a bare-groupId GAV, so no fixture/import breaks. Both the guard and the validator document how to relax to "support" if that's ever wanted.

## Tests (RED → GREEN)
- `DistributionCoordinateGuardTest` — real-DSL fixtures: bare-groupId → throws (names component + entry + reason); valid `group:artifact:ext` → no throw; `file://` URL → skipped.
- `V4WriteValidationTest` — create with blank `groupPattern` / blank `artifactPattern` → 400; well-formed coordinate → 201. (RED before the fix: blanks were accepted 201.)
- Full `build` + full server `dbTest` green.

**Post-merge:** close #349 manually (auto-close fires only on merge to the default branch `main`, not `v3`).

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -2416,6 +2416,7 @@ class ComponentManagementServiceImpl(
         config: ComponentConfigurationEntity,
         artifacts: List<MavenArtifactRequest>,
     ) {
+        artifacts.forEach { validateMavenArtifactCoordinate(it) }
         config.mavenArtifacts.clear()
         artifacts.forEachIndexed { index, req ->
             config.mavenArtifacts.add(
@@ -3341,6 +3342,25 @@ class ComponentManagementServiceImpl(
         require(value.isNotBlank()) { "package.packageType must not be blank" }
         require(value in PACKAGE_TYPE_NAMES) {
             "Invalid package.packageType: '$value'. Allowed: $PACKAGE_TYPE_NAMES"
+        }
+    }
+
+    /**
+     * A distribution Maven coordinate must carry BOTH a groupPattern and an
+     * artifactPattern — the structured-DTO equivalent of the import path's
+     * "group:artifact" requirement (parseMavenGavEntry rejects < 2 segments).
+     * A blank field was previously copied verbatim (201), silently persisting a
+     * broken coordinate — symmetric with the import silent-drop. A groupId-only
+     * coordinate (blank artifactPattern) is not supported; see TD-011 / #349
+     * (if groupId-only support is ever added, relax the artifactPattern check).
+     */
+    private fun validateMavenArtifactCoordinate(req: MavenArtifactRequest) {
+        require(req.groupPattern.isNotBlank()) {
+            "distribution.mavenArtifacts: groupPattern must not be blank"
+        }
+        require(req.artifactPattern.isNotBlank()) {
+            "distribution.mavenArtifacts: artifactPattern must not be blank " +
+                "(groupId-only coordinate '${req.groupPattern}' is not supported — see TD-011/#349)"
         }
     }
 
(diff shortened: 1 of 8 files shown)


pgorbachev:
Addressed Sonnet+Opus review in a1681792:
- **v4 embedded separators (MAJOR):** `validateMavenArtifactCoordinate` now rejects `:`/`,` in groupPattern/artifactPattern/extension/classifier — a smuggled separator corrupted the V1-compat `composeGavCsv` round-trip (import can't produce such a segment). Restores true import↔v4 symmetry.
- **ext/classifier blank normalization (MINOR):** blank → null on v4 write (matches the import parser), so `""` vs null can't skew MavenGavCollision.
- **Coverage:** added v4 separator-rejection + PATCH blank-coordinate tests.

Design note (Opus MAJOR, unchanged by request): a malformed GAV fails the **whole component's** import (same blast radius as the #387 guard), not just the one coordinate — the intended fail-loud trade. No in-repo DSL hits it; flag for the cutover runbook to confirm against prod DSL.

PR #407, opened 2026-07-04
fix(ownership): canonicalize artifact group ownership to one groupId per row (wire re-composed)

pgorbachev:
## What & why

Canonicalize artifact ownership so `component_artifact_mappings.group_pattern` holds **exactly one Maven groupId per row**. A legacy comma group-list `"a,b"` is split into one row per group (same mode / tokens / version_range). Requested by the domain owner ("one groupId = one record"), surfaced from the Portal Create-Component wizard.

Because the DB is rebuilt from the DSL on every deploy, the fix lives in **import + write logic** (not one-shot SQL) and survives redeploys.

## Design: storage-normalize + wire re-compose

The legacy v1–v3 forward wire must stay **byte-identical**. Three forward surfaces previously rendered only the **primary** (lowest `sortOrder`) mapping per range — splitting storage naively would silently **drop** non-primary groups. So a new shared helper re-composes the split rows back into the single legacy pair.

- **New** `ArtifactOwnershipGrouping` — collapses a range's mappings into maximal **contiguous** split-equivalent runs (key: `EXPLICIT`=ordered tokens, `ALL`=mode, `ALL_EXCEPT_CLAIMED`=never merge) and re-joins each run's groups by comma. Contiguity preserves author order and never over-joins.
- Re-compose wired into: `DatabaseComponentRegistryResolver.mavenArtifactParametersFor` (`/maven-artifacts`), `EntityMappers.buildEscrowModuleConfig` (escrow view / compat baseline), `ComponentCodeRenderer.writeArtifactIds` (view-as-code).
- Split at write time: `ImportServiceImpl.buildOwnershipMappings` (now for **all** modes, was `ALL_EXCEPT_CLAIMED`-only) and `ComponentManagementServiceImpl.addArtifactIds` (per-group split with a **running** `sortOrder` — request-index would collide once a request yields >1 row, and the lowest `sortOrder` is the legacy primary).

## Why it's safe

- **Semantically lossless:** `"a,b"→X` ≡ `{a→X, b→X}` — the resolver's `groupIdMatches` and the uniqueness matrix `computeOwnershipCollisions` both tokenize groups, so split vs CSV is verdict-neutral. `MavenGavCollision` (the asymmetric distribution-GAV gate) does **not** receive ownership rows (`collectDslGavRows` excludes them).
- **Wire byte-identical → no compat-comparator change.** QA data shows **zero** non-canonical `group_pattern` values in both tables (no spaces / leading / trailing / doubled commas), so split+trim+rejoin is a byte no-op on real data. `ALL_EXCEPT_CLAIMED` already normalized this way (compat-green).
- **Scope:** only the new `component_artifact_mappings` model — legacy `component_artifact_ids` is not in the v3 schema.

## Testing (TDD, RED→GREEN)

- Render re-compose guards (resolver + escrow + view-as-code): re-compose split rows; **no over-join** for distinct tokens; `ALL_EXCEPT_CLAIMED` never collapses; single legacy CSV row unchanged; independent same-shape rows merge (ALL + EXPLICIT).
- Import split + write-side split (running `sortOrder`) via focused reflection tests.
- Verified locally: server unit suite, **`:dbTest` integration gate**, compat `:unitTest`, full `./gradlew build`, and an independent adversarial review.


pgorbachev:
Addressed the Copilot review in 7d25a2d9: added `ArtifactOwnershipModeClassifier.splitGroups`, which **fails loud** (`IllegalArgumentException` → 400 on the write path) on an empty comma segment (`,`, `a,`, `a,,b`) instead of silently filtering it out. Both split sites (import `buildOwnershipMappings` and v4 `addArtifactIds`) now use it. Verified no real-data impact: QA has **zero** non-canonical `group_pattern` values in both tables, and the prior filter only ever removed *empty* segments (never a real groupId) — so this is a contract-hardening, consistent with the repo's fail-loud precedent. Added classifier + import (IMP-004) + write (WR-003) rejection tests.

PR #415, opened 2026-07-06
feat(admin): persist operational service events (SYS-060/061)

pgorbachev:
## Why
The portal Admin needs to show operational/service events — redeploys, data migrations, TeamCity resync, and (via ingest) the portal's scheduled component-validation sweep. Today migration/resync runs live only in an in-memory `AtomicReference` + logs (lost on pod restart) and redeploys aren't recorded at all. This adds a persistent, append-only journal.

## What (SYS-060 / SYS-061)
**Storage & recorder**
- `service_event` table (`V4__add_service_event.sql`; also created by Hibernate `ddl-auto` in the flyway-disabled envs). `detail` is `TEXT`+`@JdbcTypeCode(JSON)`, mirroring `audit_log`.
- `ServiceEventRecorder`: best-effort — each write in its own `REQUIRES_NEW` tx, wrapped in swallow-and-log **outside** the tx, so journaling can never roll back or crash the observed job/boot.

**Emitters (fail-loud on failure — hard requirement)**
- The 3 async jobs (components migration, history migration, TeamCity resync) record `RUNNING` at the top of the work runnable and transition it in place to `COMPLETED`/`FAILED` — **one row per run**. `triggeredBy` is captured on the caller thread (`CurrentUserResolver` / `"scheduler"`) and passed through `startAsync`.
- **Every failure path writes FAILED**: the run's `catch`, and the executor-reject path (`startAsync` catches the rejection → standalone terminal FAILED).
- `ServiceStartupListener` (`ApplicationReadyEvent`) reconciles any `RUNNING` crs row left by a dead pod → `FAILED("interrupted by restart")`, then writes a `STARTUP` marker with the build version. **Single-pod** (`replicas: 1`).

**API**
- `GET /rest/api/4/admin/service-events` — paginated (filter by type/source/status/time), IMPORT_DATA-gated like the rest of `AdminControllerV4`.
- `POST /rest/api/4/admin/service-events` — ingest for portal-sourced events (portal redeploys, validation sweeps). Shared-secret `X-Service-Event-Token` (constant-time, **fail-closed**), method-scoped `permitAll` at the filter chain above the `/rest/api/4/**` authenticated rule; accepts terminal status + `source=portal` only.

**Retention**: a `@Scheduled` daily prune deletes rows older than `components-registry.service-events.retention-days` (default 90) — scheduled, not startup-only.

## Config (service-config, per env)
- `components-registry.service-events.ingest-token` — shared secret (blank ⇒ ingest disabled, fail-closed).
[...]

PR #417, opened 2026-07-06
fix(admin): forbid re-running a completed migration (git==0 → 409)

pgorbachev:
## Why
After a full Git→DB migration, the Admin **Run migration** button stayed enabled and would re-trigger the process (reported from the stand). Re-running a completed migration is pointless (nothing left in Git; it only re-does defaults) and confusing.

## What
`POST /rest/api/4/admin/migrate` now returns **409 `{code: "migration-complete"}`** when `getMigrationStatus().git == 0` (no git-sourced components remain). Fail-loud so a direct API caller is blocked too, not just the SPA. A **partial** state (`git > 0`, e.g. after failures) still allows a retry.

Distinct `code` keeps it apart from the existing same-kind attach 409 (which carries a job body).

## Tests (TDD, red→green)
- `test:` commit — new `AdminMigrateJobControllerTest` case asserting 409 + `startAsync` never called for `git==0` (RED against the pre-fix controller).
- `fix:` commit — the guard (GREEN). `AdminControllerV4SecurityTest` (real importService, git>0) unaffected. Full `test` + `dbTest` green locally.

## Pairs with
Portal PR (disables the Run button when `git === 0` + hint): octopusden/octopus-components-management-portal (fix/block-completed-migration-rerun). Independent of the service-events work (#415).

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/AdminControllerV4.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/AdminControllerV4.kt
@@ -77,6 +77,16 @@ class AdminControllerV4(
      */
     @PostMapping("/migrate")
     fun migrate(): ResponseEntity<MigrationJobResponse> {
+        // Forbid re-running once the migration is complete: `git` is the count of DSL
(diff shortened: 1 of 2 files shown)


pgorbachev:
Folded into #415 (cherry-picked the test + fix onto the service-events branch) so the two CRS changes land in a single merge → one prod release instead of two. Superseded by #415.

PR #429, opened 2026-07-10
fix(field-override): accept exact-version range [X] as a single Maven segment

pgorbachev:
## Problem

Saving any component that carries a field-override with an **exact-version range** (a Maven "hard version" like `[1.0.49]`) failed with:

> Field-override range '[1.0.49]' must be a single Maven segment (e.g. [1.0,2.0)). Composite ranges are not accepted on POST / PATCH — split the override into multiple rows, one per segment.

`[1.0.49]` (lo==hi, both inclusive) is a valid **single** Maven segment — the simplest one — not a composite. Because the whole component re-validates on save, the pre-existing override froze **all** edits to the component.

## Root cause

`ComponentManagementServiceImpl.parseSimpleSegment` decided "single vs composite" with a regex that **requires a comma** (`^([\[(])([^,]*),([^,]*)([\])])$`), i.e. only the two-bound `[X,Y)` form. An exact-version pin has no comma, so it returned `null` and was rejected as composite. The releng `VersionRangeFactory` itself parses `[X]` fine, so syntax validation passed — the defect was purely in the composite-detector.

## Fix

- Add `EXACT_VERSION_PATTERN` (`^\[([^,\[\]()]+)]$`, square-bracket only, matching releng's hard-version rule) and parse `[X]` as `lo==hi`, both inclusive.
- Intersection / disjoint / duplicate logic is unchanged — `isIntersect` already treats `[X]` as a hard version.

## Tests (TDD, red→green)

Four cases in `V4WriteValidationTest`: accept a lone exact version, accept two disjoint exacts, reject a duplicate exact, reject an exact version inside a wider sibling range. The first three were red before the fix. Full class passes.

Independent review: clean (no bugs, no stale docs).

## Follow-up

Portal `versionRange.ts` shares the same comma-requiring regex, so the UI cannot yet *author* exact-version overrides — separate PR to follow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Field overrides now correctly accept Maven exact-version ranges, such as `[1.2.3]`.
  * Duplicate exact-version overrides and overlaps with broader ranges continue to be rejected.
  * Multiple non-overlapping exact-version overrides are supported.

* **Tests**
  * Added validation coverage for accepted, duplicate, disjoint, and overlapping exact-version ranges.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt
@@ -2703,6 +2703,15 @@ class ComponentManagementServiceImpl(
     // path inside parseSimpleSegment without a separate composite-detector.
     private val SIMPLE_SEGMENT_PATTERN = Regex("^([\\[(])([^,]*),([^,]*)([\\])])$")
 
+    // The exact-version ("hard version") form `[X]` is also a single Maven
+    // segment — the simplest one — but carries no comma, so it never matched
+    // SIMPLE_SEGMENT_PATTERN and was misclassified as composite (rejected on
+    // POST/PATCH). Maven only allows the closed `[X]` shape for a hard version:
+    // `(X)`, `[X)`, `(X]` are all invalid, so this pattern is intentionally
+    // square-bracket only. The releng VersionRangeFactory parses `[X]` as
+    // lo == hi, both inclusive — we mirror that below.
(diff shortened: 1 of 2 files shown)


pgorbachev:
Deferred follow-up (from review, low priority): `EXACT_VERSION_PATTERN` + the hard-version branch of `parseSimpleSegment` duplicate the exact-version logic already in `VersionRangePartition.parseSegment` / `SINGLE_VERSION_PATTERN`. This very bug arose because one copy supported `[X]` and the other didn't. A later refactor could unify them (delegate to a shared helper) to prevent future drift — kept out of this focused fix to avoid refactor risk.

fishinitself (approved):

PR #432, opened 2026-07-10
feat(audit): stamp a per-save correlation_id on all audit rows

pgorbachev:
Enabler for grouping the audit history by save (the CRS half; Portal grouping follows).

## Problem

A single save is one transaction that can change several entities — e.g. moving the base Java version **and** adding a per-range override in one PATCH writes two `audit_log` rows (`build.javaVersion` + `fieldOverride[build.javaVersion]`). Today those rows share nothing but a ~5 ms timestamp, so the history shows one action as two unrelated entries. The `correlation_id` column (and its index, and `AuditLogResponse.correlationId`) already exist — they were just never populated for component saves.

## Change

Populate `correlation_id` per save, transaction-bound:

- **`AuditCorrelationIdProvider`** — generates the id lazily on first use within a transaction, reuses it for the rest of that transaction, and clears the `ThreadLocal` via a `TransactionSynchronization` on completion (so a pooled request thread never leaks an id into the next save). Transaction-scoping (not request-scoping) means all rows from one `@Transactional` save — and only those — share one id.
- Stamped in the single `publishAuditEvent` choke point; persisted by `AuditEventListener`.
- **No schema change** — column/index/DTO field already present; the read path (`V4Mappers`) already returns it.

## Tests

`AuditCorrelationIdTest` (integration): two changes in one PATCH share one non-null id; a second independent save gets a different id (no cross-save leak). Both would fail before this change (every id was null).

Regression-checked the audit write path (`AuditChangeMetadataTest`, `FieldOverrideAuditTest`, `AuditLogFilterTest`) and the suites that construct the service directly (`MIG047…`, `CheapFieldFormat…`, `ArtifactGroupSplit…`) — the new dependency is defaulted so those need no wiring changes.

## Follow-up

Portal PR: group history rows by `correlationId` into one expandable transaction record (human-readable labels, inline old→new for single-field changes, `+N` overflow) — per the reviewed mockup. Legacy rows (null id) render individually, unchanged.

### DIFF
--- a/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/event/AuditCorrelationIdProvider.kt
+++ b/components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/event/AuditCorrelationIdProvider.kt
@@ -9,35 +9,60 @@ import java.util.UUID
  * Supplies the `correlation_id` shared by every audit row written during a single
  * save, so the audit history can group same-transaction changes into one record.
  *
- * The id is bound to the CURRENT TRANSACTION: generated lazily on first use,
- * reused for the rest of that transaction, and cleared when the transaction
- * completes — so a pooled request thread never leaks an id into the next save.
- * A component save is a single `@Transactional` method that publishes one
- * `AuditEvent` per changed entity (base attribute, field override, …); binding
- * to the transaction rather than the request means all of those rows — and only
- * those — share one id, even if a request were ever to span several transactions.
+ * The id is bound to the CURRENT TRANSACTION as a [TransactionSynchronizationManager]
+ * resource: generated lazily on first use, reused for the rest of that transaction,
+ * and unbound when the transaction completes. Binding it as a managed resource (not
+ * a raw `ThreadLocal`) means Spring suspends/resumes it correctly, so a nested
+ * `REQUIRES_NEW` transaction gets its OWN id and the outer id is restored afterwards
+ * — a plain `ThreadLocal` would leak the outer id into the inner transaction.
  *
- * Called outside an active transaction (should not happen for saves, but e.g.
- * background paths that publish audit events directly), it returns a fresh id
- * each call: no grouping, but also no `ThreadLocal` leak.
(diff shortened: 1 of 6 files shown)


fishinitself:
Reviewed this — implementation is solid (correctness-focused passes found zero bugs; transaction lifecycle, thread-safety, and the defaulted-constructor-param trick for legacy unit tests all check out). Two things worth considering before merge:

**1. Layer placement of the correlation-id call.** `AuditEventListener.handleAuditEvent` (`@TransactionalEventListener(phase = BEFORE_COMMIT)`) is already the single funnel every `AuditEvent` passes through, still inside the originating transaction. `AuditCorrelationIdProvider.current()` is a pure function of "which transaction is active," so calling it from the listener would return the identical id as calling it from `ComponentManagementServiceImpl.publishAuditEvent`. As written, any future service that publishes `AuditEvent` must remember to inject `AuditCorrelationIdProvider` and call `.current()` itself — forgetting silently produces an ungrouped (null-correlationId) row instead of failing loudly. Moving the call into the listener would make grouping automatic for every current and future publisher, closing that gap structurally.

**2. No numbered requirement was created for this feature.** Per `AGENTS.md`: *"Every new feature → requirements first, then implementation"*, and test methods covering a numbered requirement must embed the ID (`SYS-xxx`) in the method name / `@DisplayName`. I don't see a corresponding entry in `docs/registry/requirements-common.md`, and neither `AuditCorrelationIdTest.kt` nor `AuditCorrelationIdProviderTest.kt` references a requirement ID. Might be an intentional exception for this kind of enabler PR, but flagging since the rule is unconditional as written.

(For the record, I also chased down a candidate finding that the manual `suspend()`/`resume()` logic in `AuditCorrelationIdProvider` duplicates Spring's own resource-suspension — verified against `AbstractPlatformTransactionManager` internals and it's a non-issue: the framework only auto-suspends resources bound by the transaction manager itself, not arbitrary application-level keys like `RESOURCE_KEY`, so this manual handling is necessary, not redundant.)

fishinitself (approved):

rodemark (approved):

PR #434, opened 2026-07-13
feat(security): grant edit to manager of component owner 

fishinitself:
## Summary
- Component owner's manager can now edit the component: `PermissionEvaluator.canEditComponent` consults a new `EmployeeDirectoryService.getManager()` (backed by the employee-service client's `getManager` API), fail-open/fail-closed consistent with the existing owner/RM/SC checks.
- Bumps `octopus-employee-service` to 2.0.38 to pick up the `getManager` client API.
- That version bump added new methods (`getConnectTimeoutMillis`/`getReadTimeoutMillis`/`getConnectionTtlMillis`) to `EmployeeServiceClientParametersProvider`, which broke compilation of the legacy CI validation task's Groovy anonymous implementation of that interface — fixed by implementing the three methods there with standard HTTP client timeout defaults (5s connect / 10s read / 5min TTL).
- Fixed detekt (`ClassOrdering`) and ktlint (chain-wrapping, import-ordering, function-signature) violations in the test/source files, surfaced by the `[1.0] Compile & UT` static-analysis gate that was enabled after this branch was forked from main.

## Summary of Release Notes

* **New Features**
  * Expanded editing rights: a component owner’s manager can now PATCH/edit owned components (SYS-063).
* **Bug Fixes**
  * Improved reliability of employee-directory lookups with explicit connection, read timeout, and connection TTL settings.
* **Tests**
  * Added authorization test coverage for manager-allowed, manager-denied, and “no manager” scenarios.
* **Documentation**
  * Updated permission requirements and technical documentation to reflect the new manager-based edit rule.


gradle.properties
```diff
@@ -22,7 +22,7 @@ org.gradle.configuration-cache=false
 # ow deps
 octopus-external-systems-clients.version=2.0.90
 octopus-release-management.version=2.0.4
-octopus-employee-service.version=2.0.37
+octopus-employee-service.version=2.0.38
```
pgorbachev, line 25: This bump has a runtime side effect that's easy to miss: the three new `EmployeeServiceClientParametersProvider` methods have **default implementations** in 2.0.38 — `getConnectTimeoutMillis()` and `getReadTimeoutMillis()` both default to **300000 ms**, `getConnectionTtlMillis()` to -1 (I checked the interface bytecode in the published jar). And `ClassicEmployeeServiceClient` really does feed them into `feign.Request.Options`, so they're live.

The PR overrides them in the Groovy validation task, but the server-side `EmployeeServiceConfig.kt` anonymous provider was left as is — so the **runtime** client now runs with 5-minute connect/read timeouts. Before this PR that was mostly tolerable (the client only sat on fail-open validation/search paths), but `canEditComponent` now calls it inside `@PreAuthorize` and inside `withCanEdit` on every detail GET. A hung employee-service would pin servlet threads for up to 5 minutes per request — a handful of concurrent viewers and the pool is gone.

Suggestion: override the three methods in `EmployeeServiceConfig` as well, ideally wired through `EmployeeServiceProperties` (with 5s / 10s / 5min as defaults) rather than hardcoding a second copy of the numbers.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/security/PermissionEvaluator.kt
```diff
@@ -60,9 +63,13 @@ class PermissionEvaluator(
...
             // owner, empty RM, empty SC) — falls through to deny (owner-less is then admin-only,
             // reachable only via the EDIT_ANY_COMPONENT bypass handled above).
-            if (matches(username, repository.findComponentOwnerById(id))) return true
+            val owner = repository.findComponentOwnerById(id)
+            if (matches(username, owner)) return true
             if (repository.findReleaseManagerUsernames(id).any { matches(username, it) }) return true
             if (repository.findSecurityChampionUsernames(id).any { matches(username, it) }) return true
+            // the manager of the componentOwner may also edit.
+            val ownerTrimmed = owner?.trim()
+            if (!ownerTrimmed.isNullOrEmpty() && matches(username, employeeDirectory.getManager(ownerTrimmed))) return true
```
pgorbachev, line 72: This puts a synchronous HTTP call on a pretty hot path: `withCanEdit` stamps `canEdit` onto **every** detail response, so for a regular viewer who isn't owner/RM/SC (the common case) every detail GET now makes a remote employee-service call. A manager doing a PATCH triggers it twice — once in `@PreAuthorize`, once when stamping the response. There's no caching, so even with sane timeouts a degraded employee-service adds its full latency to every detail view.

A short-TTL cache keyed by the owner username inside `EmployeeDirectoryService.getManager` (Caffeine, 1–5 min) would basically make this free — there are few distinct owners and they repeat heavily, and a minute of staleness is fine for a permission of this kind.

Also, the KDoc above ("only when they are listed on it as its `componentOwner`, a `releaseManager`, or a `securityChampion`" + resolution-order step 5) still describes the old rule — worth updating in this PR since it's the reference description of the gate.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/EmployeeDirectoryService.kt
```diff
@@ -148,6 +148,23 @@ class EmployeeDirectoryService(
     /** Coarse health enum for callers that don't need the failure detail. */
     fun probe(): IntegrationHealth = probeHealth().health
 
+    /**
+     * Returns the manager's sAMAccountName for [ownerUsername], or null when the
+     * user has no manager, is not found, or the service is unavailable (fail-open).
```
pgorbachev: Nit on the wording: "(fail-open)" is copied from the lookup methods above, but it means the opposite here. For a permission *grant*, returning `null` on failure **denies** — that's fail-closed, and it's exactly what we want. The behavior is right; I'd just reword the doc so the next person doing a security pass doesn't stumble over it (e.g. "…or the service is unavailable — so a directory failure can only deny the manager grant, never allow").

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/EmployeeDirectoryService.kt
```diff
@@ -148,6 +148,23 @@ class EmployeeDirectoryService(
...
+     */
+    @Suppress("TooGenericExceptionCaught")
+    fun getManager(ownerUsername: String): String? {
+        val client = clientProvider.getIfAvailable() ?: return null
+        return try {
+            client.getManager(ownerUsername).manager
+        } catch (e: NotFoundException) {
```
pgorbachev: Tiny one: `isActive` above logs a `debug` on `NotFoundException`, this catch is silent. A matching `log.debug("employee-service: getManager('{}') — owner not found", ownerUsername)` would keep the two consistent and help when someone wonders why a manager grant didn't kick in. Not blocking.

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/controller/ComponentOwnershipEditSecurityTest.kt
```diff
@@ -320,14 +339,37 @@ class ComponentOwnershipEditSecurityTest {
...
         performGet(id, jwt)
             .andExpect(status().isOk)
-            .andReturn().response.contentAsString
+            .andReturn()
+            .response.contentAsString
+
+    // --- OCTOPUS-2191: manager of componentOwner ----------------------------------
+
+    @Test
+    @DisplayName("manager of componentOwner can PATCH the component (OCTOPUS-2191)")
```
pgorbachev: Traceability per AGENTS.md: a new feature is supposed to get a numbered requirement in `docs/registry/requirements-common.md` first (with acceptance criteria), and the tests then carry that ID in the method name and `@DisplayName` ("SYS-NNN: …"). `OCTOPUS-2191` is the Jira key, not a requirement ID — right now this feature isn't findable from the requirements ledger at all, and `technical-design.md` still describes the edit gate as owner/RM/SC/admin.

Suggestion: add a SYS-NNN entry for the manager grant, rename the five new tests (here and in `PermissionEvaluatorTest`) to carry it, and touch up the technical-design section on the edit gate.

While writing the acceptance criteria, one behavior is worth an explicit decision: `GET /{idOrName}/editors` still returns owner + RM + SC only, so the manager can edit but won't appear in the Portal's "who can edit" list. If that's intended (derived grant, not a listed editor), the requirement should say so; if not, that endpoint needs the same lookup.

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/security/PermissionEvaluatorTest.kt
```diff
@@ -193,4 +201,45 @@ class PermissionEvaluatorTest {
         assertFalse(evaluator.canEditComponent(id.toString()))
         verify(repo, never()).findComponentOwnerById(id)
     }
+
+    // --- OCTOPUS-2191: manager of componentOwner ----------------------------------
+
+    private fun evaluatorWithManager(
```
pgorbachev, line 207: One case missing from the matrix: the transport-failure path. All three new tests exercise a *responding* directory (manager / other-manager / null); nothing covers `client.getManager` throwing a non-NotFound exception, which is the path that guarantees an employee-service outage denies rather than 500s or (worse) grants. It's cheap here — mock the client to throw a `RuntimeException` and assert deny — and it pins down the fail-closed contract that the `getManager` catch block currently provides.

component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/ComponentRegistryValidationTask.groovy
```diff
@@ -220,6 +220,21 @@ class ComponentRegistryValidationTask {
                 return 5000
             }
 
+            @Override
+            int getConnectTimeoutMillis() {
```
pgorbachev, line 224: The values are sensible for a CI-only task, so no objection to the numbers themselves — but if `EmployeeServiceConfig` gets its own timeout overrides (see my comment on `gradle.properties`), we'll have two independent copies of 5000/10000/300000 that can drift. Named constants here, or the same defaults surfaced in one place, would keep them honest. Nit, not blocking.

pgorbachev (commented):
Overall this looks solid — the failure direction is right everywhere (a directory failure can only deny the manager grant, never allow), the check is correctly ordered last after the cheap DB lookups, and the test coverage of allow/deny/null-manager is good.

Two things I'd want addressed before merge though: the employee-service client's new 5-minute default timeouts now sit on the authorization path of every detail GET (details inline on `gradle.properties`), and the feature is missing its SYS-NNN requirement / design-doc update per AGENTS.md. The rest is smaller — details inline.

pgorbachev (changes requested):
Marking as changes-requested per my review above (https://github.com/octopusden/octopus-components-registry-service/pull/434#pullrequestreview-4685768681): the blocking item is the runtime employee-service client keeping the 2.0.38 default 5-minute connect/read timeouts on the new authorization path (see the `gradle.properties` comment), plus the missing SYS-NNN requirement / design-doc update. The rest of the inline comments are non-blocking.

fishinitself:
Pushed `6a77ebd` addressing both blocking items:

- **Runtime timeouts**: `EmployeeServiceConfig`'s live client now overrides `getConnectTimeoutMillis`/`getReadTimeoutMillis`/`getConnectionTtlMillis` via new configurable `EmployeeServiceProperties` fields (5s/10s/5min defaults) instead of inheriting the library's 5-minute defaults on the `canEditComponent` hot path.
- **Traceability**: added `SYS-062` to `requirements-common.md` (acceptance criteria + test method), updated `technical-design.md` §6.3, and renamed the 5 manager-of-owner tests to carry `SYS-062` in method name + `@DisplayName`.

The non-blocking nits (owner-manager caching, `EmployeeDirectoryService` doc wording, missing transport-failure test, shared timeout constants between the Groovy task and `EmployeeServiceConfig`) are left as follow-ups rather than silently dropped — happy to take those in this PR too if you'd rather not split it.

pgorbachev:
The fixup itself looks right — the timeout overrides landed exactly where they should and the traceability is in place — but two things happened on main in the meantime:

1. **The PR is now CONFLICTING with main** (`requirements-common.md`), which is also why CI looks stuck: GitHub can't build the merge ref for a conflicting PR, so the `pull_request`-triggered Merge Gate never starts. It's not a flaky runner — no run will appear until the conflict is resolved.

2. **SYS-062 is taken.** #433 merged earlier today and claimed `SYS-062: User feedback / report-a-problem` in `requirements-common.md`. The manager-grant requirement needs the next free number (SYS-063 as of right now) — in `requirements-common.md`, the two mentions in `technical-design.md` §6.3, and the five test names / `@DisplayName`s in `PermissionEvaluatorTest` and `ComponentOwnershipEditSecurityTest`.

So: rebase on main, renumber, push — CI should then trigger normally, and I'll flip my review.

fishinitself:
Done — pushed `9c74f74`:

- Rebased/merged `main` (picked up #432, #433, #435).
- Renumbered the manager-grant requirement to **SYS-063** (SYS-062 was taken by #433's feedback feature) across `requirements-common.md`, `technical-design.md` §6.3 (2 mentions), and all 5 test method names / `@DisplayName`s in `PermissionEvaluatorTest` and `ComponentOwnershipEditSecurityTest`.

Verified locally post-merge: compile, detekt, ktlintCheck, and the full `components-registry-service-server` test suite all pass. Should be unblocked for CI now that the merge conflict is resolved.

fishinitself:
Pushed `b070757` — resolved a second merge conflict from `main`'s #428 (mechanical `ktlintFormat` pass across all modules, no behavior change). Both conflicting spots (`EmployeeServiceConfig.kt`, `ComponentOwnershipEditSecurityTest.kt`) were trivial: our additions vs. unrelated reformatting on adjacent lines.

GitHub now reports `mergeable: MERGEABLE` (conflict cleared). Verified locally post-merge: full-repo compile, `detekt`, `ktlintCheck`, and the full `components-registry-service-server` test suite all pass. CI checks are running now.

pgorbachev:
I think SYS-063 should also update the informational `GET /rest/api/4/components/{idOrName}/editors` contract.

Once the component owner's manager can edit, the Portal's **"Who can edit this component"** panel becomes misleading if this endpoint deliberately omits that manager. The endpoint should return the manager grant as well.

Rather than adding another one-off field, I suggest exposing structured edit grants (while preserving the current fields for backward compatibility), for example:

```json
{
  "editGrants": [
    {
      "subjectType": "USER",
      "subjectId": "owner-user",
      "reason": "COMPONENT_OWNER"
    },
    {
      "subjectType": "USER",
      "subjectId": "manager-user",
      "reason": "OWNER_MANAGER"
    },
    {
      "subjectType": "USER",
      "subjectId": "rm-user",
      "reason": "RELEASE_MANAGER"
    },
    {
      "subjectType": "GROUP",
      "subjectId": "team-payments",
      "reason": "COMPONENT_TEAM"
    },
    {
      "subjectType": "PERMISSION",
      "subjectId": "EDIT_ANY_COMPONENT",
      "reason": "GLOBAL_PERMISSION"
    }
  ]
}
```

This gives the UI enough information to show both **who** has edit access and **why**, and it avoids another contract change if team/group-based grants are added later. Groups or global permissions should be represented as grants rather than expanded into every member.

Ideally CRS should build this response from the same grant-resolution logic used by `canEditComponent`; the Portal should not call employee-service directly. The current SYS-063 wording that the manager grant is intentionally unlisted should be updated accordingly.

fishinitself:
Good point on `editGrants` — the "who and why" framing and the group/team-friendly shape are genuinely better long-term. We're going to ship the simpler `manager` field for now (already pushed in `230f508`) rather than build the general grants API in this PR, for a few reasons:

- There's no group/team-based grant anywhere in this codebase today — every grant type resolves to an individual username (owner/RM/SC/manager) or a role check (admin). Building `subjectType: GROUP` support now means designing for a shape we don't have a real second use case for yet.
- `canEditComponent` today is a short-circuiting boolean (returns on the first match). Emitting a full `editGrants` list needs the opposite — evaluate and collect every matching branch — so "reuse the same grant-resolution logic" is really a second code path to keep in sync with the first, not a reuse.
- It's a meaningfully bigger scope than this PR's stated goal (new DTO + reason enum + Portal UI to render it), on top of two rounds of review already.

We'll treat `editGrants` as a follow-up once there's an actual second grant type (e.g. team/group ownership) that needs it — happy to open a ticket for it now if you'd like it tracked rather than left implicit. SYS-063's wording has already been updated to say the manager grant IS enumerated on `/editors` (no longer "intentionally unlisted"), per your original comment.

pgorbachev (commented):
Two remaining issues before approval:

- **[P1] Move the employee-service lookup outside the DB transaction.** In `ComponentManagementServiceImpl.getEditors`, `findByIdOrName` and the lazy editor collections acquire DB resources, then `employeeDirectory.getManager()` may wait for the remote read timeout while `@Transactional(readOnly = true)` remains open. During an employee-service outage, failures are intentionally not cached, so concurrent `/editors` requests can each hold a DB connection for up to 10 seconds and exhaust the pool. A non-transactional facade can first obtain the explicit editors through this transactional service and then resolve/copy `manager` after the transaction has completed.

- **[P2] Complete the SYS-063 requirement metadata.** Add SYS-063 to the Summary Table and include explicit `**Priority:**`, `**Test layer:**`, and `**Status:** ✅ Tested` fields in the SYS-063 section. The test methods are already listed, but without the status/index entry the requirement remains incomplete according to the repository traceability rules.

fishinitself:
Pushed `a7044ca` addressing both:

- **[P1] Transaction boundary**: `getEditors` no longer holds a DB connection across the employee-service lookup. The class carries a class-level `@Transactional`, so I explicitly override `getEditors` with `propagation = Propagation.NOT_SUPPORTED` — just dropping the method-level annotation would not have been enough. The DB read (entity + LAZY `releaseManagers`/`securityChampions`) now runs in its own short-lived transaction via an explicit `TransactionTemplate` (added `PlatformTransactionManager` to the constructor), which closes before `employeeDirectory.getManager()` runs — same explicit-template-over-self-invocation pattern already used by `ServiceEventRecorderImpl`/`TeamcitySyncService` in this codebase.
- **[P2] SYS-063 metadata**: added `**Priority:**`/`**Test layer:**`/`**Status:** ✅ Tested` fields and a Summary Table row in `requirements-common.md`.

Verified locally: compile, detekt, ktlintCheck all clean; full fast suite (778 tests) + targeted `dbTest` runs of the two affected integration test classes (26 tests, including the manager-present/manager-null `/editors` cases) all pass.

fishinitself:
Pushed `deb8cb6` — merged `main` (picked up #436 TeamCity project/version-line rework, #437 CI restructuring).

Renumbered our requirement to **SYS-064** — #436 independently claimed `SYS-063` for its own TeamCity requirement. Updated across `requirements-common.md` (prose + Summary Table), `technical-design.md`, `ADR-004`, `functional-spec.md`, and all 8 test names/`@DisplayName`s in `PermissionEvaluatorTest`/`ComponentOwnershipEditSecurityTest`.

`mergeable: MERGEABLE` now (conflict cleared). Verified locally: compile, detekt, ktlintCheck all pass; full fast suite (798 tests) + targeted `dbTest` runs of both our integration test classes and #436's own TeamCity test class (30 tests total) all pass.

pgorbachev (approved):

rodemark (approved):

fishinitself:
### Review: two issues worth addressing before merge

**1. `getEditors` triggers an unauthenticated outbound employee-service call — no anonymous guard**

`components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/service/impl/ComponentManagementServiceImpl.kt:479`

`PermissionEvaluator.canEditComponent` short-circuits at line 57 on `username.isEmpty() || username == ANONYMOUS_USER`, **before** reaching the new `employeeDirectory.getManager(...)` call — so the `canEdit`-stamping path correctly never calls employee-service for unauthenticated callers.

`getEditors` has no equivalent guard: `manager` is resolved for every caller. And `WebSecurityConfig.kt:89-90` makes `GET /rest/api/4/components/**` `permitAll()`, with `ROLE_ANONYMOUS` holding `ACCESS_COMPONENTS` — so `@PreAuthorize("...hasPermission('ACCESS_COMPONENTS')")` on `/editors` passes **without a JWT**.

Scenario: an unauthenticated caller walks `/rest/api/4/components/{id}/editors` across N components with distinct owners → N outbound employee-service calls. Resolved answers cache for 2 minutes, but lookup failures are deliberately **not** cached (`EmployeeDirectoryService.kt:190-193` returns before `managerCache.put`). So while employee-service is degraded, every such request blocks for the full connect + read budget (5 s + 10 s) holding a servlet thread, with no per-caller bound — externally reachable thread-pool exhaustion, unauthenticated.

Secondary: the `manager` field publishes manager → report org-chart relationships to anonymous callers. `componentOwner` alone did not expose that relationship.

Suggested fix: mirror the anonymous/blank-username short-circuit in `getEditors` (or only populate `manager` for authenticated callers), and consider a short negative cache or circuit breaker for failed resolves — useful on its own, since today there is no caching at all for the duration of an outage.

**2. Read timeout of 10 s can hard-fail the CI validation task**

`component-resolver-core/src/main/groovy/org.octopusden/octopus/escrow/configuration/validation/ComponentRegistryValidationTask.groovy:229`

`getReadTimeoutMillis()` drops from the library default (300 000 ms) to 10 000 ms. But the caller, `findErrors` (lines 250-265), catches **only** `NotFoundException`:

```groovy
try {
    def employee = employeeServiceClient.getEmployee(componentOwner)
    ...
} catch (NotFoundException e) { ... }
```

Scenario: during a validation build employee-service takes longer than 10 s for a single owner. The client throws a read/socket-timeout exception, which is not `NotFoundException`, so it escapes the `collect` closure, propagates out of `validate()`, and fails the entire registry validation task. With the previous 300 s default the same slow response would simply have completed.

The runtime server path is fine — `EmployeeDirectoryService` catches `Exception` broadly and degrades to `UNAVAILABLE`/`null`. Only this Groovy task lacks a catch-all.

Suggested fix: either widen the catch in `findErrors` (a transport timeout is not a data-validation error and should not fail the build), or keep a generous read timeout in this provider. The rationale in `EmployeeServiceProperties.kt:43-50` — a degraded employee-service pinning servlet threads inside `@PreAuthorize` — does not apply to a batch CI task, which has no servlet threads to protect.

fishinitself:
Note on the comment above: this review was run after the squash merge, so both issues are live in `main` today — they need a follow-up fix, not a pre-merge change.

Line references against current `main` (one correction to the comment above):

- Finding 1, the unguarded manager lookup: `ComponentManagementServiceImpl.kt:510` (the comment above cites 479, which was the line number in my local branch).
- `PermissionEvaluator.kt:57` (the anonymous short-circuit that `getEditors` lacks) and `PermissionEvaluator.kt:72` (the guarded manager lookup) — unchanged.
- `WebSecurityConfig.kt:89` (`permitAll()` on `GET /rest/api/4/components/**`) — unchanged.
- `EmployeeDirectoryService.kt:181-194` (`getManager`; the `return null` at 191-192 is the uncached-failure path, before `managerCache.put` at 194) — unchanged.
- Finding 2: `ComponentRegistryValidationTask.groovy:229` — unchanged.

PR #436, opened 2026-07-14
Adjust TC project for components implementation

adellinekania:
## Summary by CodeRabbit

* **New Features**
  * TeamCity synchronization now supports multiple release lines per component using `PROJECT_VERSION`.
  * Archived and fully paused projects are excluded from synchronization.
  * Component details now include the nullable TeamCity project version.
  * Sync results report the number of dropped release lines.
  * Matching now resolves parameter references and applies improved project-selection rules.

* **Bug Fixes**
  * Improved handling of ambiguous TeamCity matches and paused release builds.
  * TeamCity project links now use the correct project URL format.

* **Documentation**
  * Updated API changelog, requirements, schema, and technical design documentation.


components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/dto/v4/ComponentConfigurationDtos.kt
```diff
@@ -404,6 +404,7 @@ data class TeamcityProjectResponse(
     val projectId: String,
     val projectUrl: String? = null,
     val sortOrder: Int,
+    val projectVersion: String? = null,
```
pgorbachev, line 409: `projectVersion` is exposed in the response, but `TeamcityProjectRequest` contains only `projectId`. During PATCH, `ComponentManagementServiceImpl` clears the collection and recreates every row, so all submitted projects get `projectVersion = null`.

This means a normal read-modify-write operation erases the release-line metadata until the next TeamCity sync.

Because `projectVersion` is sync-owned, it does not necessarily need to become writable. However, replacement should preserve the existing `projectVersion` for an unchanged `projectId`, or otherwise avoid clearing sync-owned metadata.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/TeamcityClientConfig.kt
```diff
@@ -134,23 +136,98 @@ internal fun mapTcProjectsToComponentMatches(
...
+    project: ExternalTeamcityProject,
+    projectParams: Map<String, String>,
+): String? {
+    resolveParam(PROJECT_VERSION_PARAM, projectParams)?.let { return it }
+
+    return project.buildTypes
+        ?.buildTypes
+        ?.asSequence()
+        ?.filter { it.paused != true }
+        ?.firstNotNullOfOrNull { bt ->
```
pgorbachev: When the project itself has no `PROJECT_VERSION`, this fallback selects the first active build type declaring it. The order comes from the TeamCity API and is not guaranteed to be stable.

If two active build types declare different versions, repeated syncs may assign the same project to different release lines without any TeamCity configuration change.

Please make the selection deterministic, for example by sorting eligible build types by `id` before scanning, or treat conflicting values as ambiguous and skip the project.

components-registry-service-server/src/main/resources/openapi/v4.json
```diff
@@ -2583,6 +2583,9 @@
           "projectUrl" : {
             "type" : "string"
           },
+          "projectVersion" : {
```
pgorbachev, line 2587: The Kotlin response property is nullable and the API documentation says it may be null, but the OpenAPI 3.0 schema permits only a string value.

Spring may emit `"projectVersion": null`, which is invalid against this schema and can be rejected by strict validators or generated clients.

Please declare nullability at the DTO/schema-generation source and regenerate `v4.json`, rather than editing only the generated file.

pgorbachev (changes requested):
The overall direction looks reasonable as an intermediate model.

A separate `VersionLine` domain entity would be premature because CRS does not yet have canonical version-line data or lifecycle rules. For now, storing `projectVersion` alongside the component-to-TeamCity-project association is a pragmatic denormalized projection of TeamCity state.

Please make this transitional semantic explicit:

- `projectVersion` is nullable and derived from TeamCity;
- TeamCity sync owns and refreshes it;
- it is a discriminator used during reconciliation, not a CRS source of truth;
- the model should remain migratable to an explicit `Component -> VersionLine -> TeamCityProject` relationship later.

Before merging, I think we should fix the loss of this metadata during v4 updates and make the build-type fallback deterministic.

Also, according to the repository rules, this new reconciliation behavior needs its own numbered SYS requirement and test traceability. SYS-051 currently covers only the decision not to produce audit rows.

pgorbachev (approved):
Previous review comments are addressed. The transitional TeamCity-derived version-line model is acceptable for this stage, and the existing associations can be regenerated by TeamCity resync.

gnh374 (approved):

PR #443, opened 2026-07-20
TeamCity Validation

adellinekania:
## Description

Adds a validation feature that inspects every registered component's TeamCity projects to understand how each one builds and which Java/Maven versions it uses — flagging Java 1.8 in particular, so we can drive its retirement. Findings are stored and exposed through read APIs for both the component detail view and an admin dashboard.

## Design

Split into a **pure validation library** and **server-side IO**:

- `component-validation` — a new pure-Kotlin Gradle module (no Spring, no DB, no TeamCity client, no IO). Projects in, `List<ValidationResult>` out. Generic `core` (`Validator` / `ValidatorSuite` / `ValidationResult`) with TeamCity as the first feature; built to serve other services later.
- The **server** owns both edges: it fetches projects from TeamCity, maps the client DTOs to the module model, runs the suite, and persists results. Dependency flows one way: `server → validation module`.

## Validators (per project)

- `ATTACHED_TO_BUILD_TEMPLATE` — at least one build config inherits a known build template (`CDGradleBuild` / `CDJavaMavenBuild`).
- `OVERRIDES_DEFAULT_BUILD_STEP` — whether a template's default build step is overridden.
- `HAS_CUSTOM_JAVA_BUILD_STEP` / `HAS_CUSTOM_MAVEN_BUILD_STEP` — custom runner steps outside the
  template.
- `USES_OLD_JAVA_VERSION` — detects Java 1.8 usage.
- Multiple-version detection for Java/Maven within a project.

## Server integration

- **Fetch**: `EnrichedTcProjectFetcher` pulls one enriched project at a time by project id (build configs + steps + template ancestry + parameters), with a short-TTL cache; decoupled from sync.
- **Persist** (`V7__add_teamcity_validation.sql`): `teamcity_validation (project_id, type, status, message, updated_at)`, PK `(project_id, type)`. Stores **only WARNING/ERROR**; **latest-only** — each run replaces that project's rows, and stale projects are pruned.
- **Triggers**: REST (`POST /admin/teamcity-validation`) and automatically after a TeamCity sync completes (`TeamcitySyncCompletedEvent` → `TeamcityValidationSyncListener`).
- **Concurrency**: TC sync and TC validation are mutually exclusive via the shared single-pod `MigrationLifecycleGate` (new `TC_VALIDATION` job kind); runs are one-at-a-time.
- New `TEAMCITY_VALIDATION` service-event type (SYS-060 journal, SYSTEM category).

## Read APIs

- **Component detail** — generic `{status, message}` validations embedded in `TeamcityProjectResponse`.
- **Admin dashboard** (`GET /rest/api/4/admin/teamcity-validations` + `/summary`) — component- centric: filter which components fail, count how many violate each check (distinct-component counts). IMPORT_DATA-gated like the other admin read APIs.

## Refactor

- Split the `teamcity` package into `common` / `sync` / `validation` to remove a dependency cycle; moved `TeamcityProperties` and the sync classes accordingly (mostly renames).

## Tests

- Full unit coverage of the validation module (validators, resolvers, model).
- Server: `TeamcityValidationQueryServiceTest` (component join, shared-project fan-out, type filter, distinct-component summary); updated sync tests and the service-event category test for the new members.

## Follow-ups (tracked)

- **TD-016** — version resolvers match against a hardcoded token list, not arbitrary versions.

## Upgrade Notes
- This release **must be deployed together with the corresponding components-management-portal release** that supports the new validation contract (`TeamcityProjectResponse.validations` + the `teamcity-validations` admin endpoints. Do not roll this service out on its own.
- New properties under `teamcity.validation.*` (`TeamcityValidationProperties`, prefix bound via `@ConfigurationProperties`). All have defaults so the service still boots without them, but the feature needs the real values for your TeamCity instance to be useful:


components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationService.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /** Full per-project replace, in its own transaction, so the flip is atomic for readers. */
+    private fun replaceFindings(
+        projectId: String,
+        issues: List<org.octopusden.octopus.validation.core.ValidationResult>,
+    ) {
+        val now = Instant.now()
+        transactionTemplate.executeWithoutResult {
+            teamcityValidationRepository.deleteByProjectId(projectId)
```
pgorbachev, line 93: [P1] `deleteByProjectId` is a derived delete, so Spring Data loads matching entities and calls `EntityManager.remove`; the deleted instances remain in the persistence context until flush. The replacement entities have assigned composite IDs and no version field, so `saveAll` treats them as existing and calls `merge`.

When a finding type survives between runs, Hibernate encounters a deleted managed instance with the same ID and can throw `ObjectDeletedException` or hit insert-before-delete PK ordering. The per-project catch then quietly retains the old findings and increments `failed`.

Please use a bulk delete with explicit flush/clear semantics, or update/diff the existing rows. Add a DB-backed test that performs two consecutive replacements with the same `(projectId, type)`.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationService.kt
```diff
@@ -0,0 +1,126 @@
...
+    private val transactionTemplate: TransactionTemplate,
+) {
+    private val log = KotlinLogging.logger {}
+
+    // The module suite is stateless given a catalog — build once.
+    private val validators = TeamCityValidators(templateCatalog)
+
+    @Suppress("TooGenericExceptionCaught")
+    fun validate(): TeamcityValidationResult {
+        val knownProjectIds = teamcityProjectRepository.findDistinctProjectIdsSafely()
```
pgorbachev: [P2] `teamcity_project` is effectively append-only: sync replaces `version_line` rows but never removes orphaned project rows. Therefore this query returns every TeamCity project ever seen rather than projects currently linked to components.

This also makes `removeStaleProjects` ineffective, because an unlinked project remains in `knownProjectIds` forever. Its findings become invisible through the component join, but validation continues requesting it from TeamCity on every run.

Please derive the scope from distinct projects referenced by current `version_line` rows, or introduce explicit orphan cleanup with matching lifecycle semantics.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationService.kt
```diff
@@ -0,0 +1,126 @@
...
+                        )
+                    },
+                )
+            }
+        }
+    }
+
+    /** Delete findings for projects no longer known. Skips when the known set is empty (never wipes all). */
+    private fun removeStaleProjects(knownProjectIds: Set<String>): Int {
+        if (knownProjectIds.isEmpty()) {
```
pgorbachev: [P2] An empty scope is a valid state when the last TeamCity project is unlinked. In that case every stored validation row is stale, but this guard retains all rows indefinitely.

If this is intended to protect against a failed scope query, that failure should be represented explicitly rather than treating a successfully loaded empty set as an error. Otherwise the advertised latest-only/stale-pruning contract is not satisfied.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/dto/v4/ValidationResponse.kt
```diff
@@ -0,0 +1,7 @@
+package org.octopusden.octopus.components.registry.server.dto.v4
+
+/** Minimal validation finding for the UI: status + message. */
+data class ValidationResponse(
+    val status: String,
```
pgorbachev, line 13: [P2] The stored finding and admin API expose a stable validation `type`, but component detail drops it and returns only a human-readable message. The Portal would have to parse presentation text to identify, group, localize, or link a finding.

Please include at least `type` in this DTO. Including `updatedAt` would also let consumers identify retained findings after a failed validation attempt.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationSyncListener.kt
```diff
@@ -0,0 +1,34 @@
...
+class TeamcityValidationSyncListener(
+    private val teamcityValidationJobService: TeamcityValidationJobService,
+) {
+    private val log = KotlinLogging.logger {}
+
+    @EventListener
+    @Suppress("TooGenericExceptionCaught")
+    fun onSyncCompleted(event: TeamcitySyncCompletedEvent) {
+        try {
+            val outcome = teamcityValidationJobService.startAsync(POST_SYNC_TRIGGER)
```
pgorbachev, line 26: [P2] This listener starts validation unconditionally: `teamcity.validation.enabled` is never consulted, while both default build-step IDs default to blank.

With those defaults, inherited default steps cannot be located, so checks such as default-step override and old-Java detection can persist false clean results after every sync.

Please either gate automatic execution on `enabled` and validate the required IDs, or fail fast when the feature is enabled with incomplete configuration. If no scheduler/gate is intended, remove the unused `enabled` property.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/TeamcityValidationControllerV4.kt
```diff
@@ -0,0 +1,37 @@
...
+@PreAuthorize("@permissionEvaluator.canImport()")
+@ConditionalOnDatabaseEnabled
+@RestController
+@RequestMapping("rest/api/4/admin/teamcity-validations")
+class TeamcityValidationControllerV4(
+    private val teamcityValidationQueryService: TeamcityValidationQueryService,
+) {
+    /** Component-centric findings; optional filters: validation `type`, `status`, `component` (name substring). */
+    @GetMapping
+    fun list(
```
pgorbachev, line 33: [P2] Adding another controller method named `list` changes unrelated generated OpenAPI operation IDs: service events move from `list` to `list_1`, and feedback moves from `list_1` to `list_2`.

This creates generated-client churn and can break consumers keyed by operation ID. Please assign an explicit stable ID such as `listTeamcityValidations` with `@Operation(operationId = ...)`. Ideally the existing list endpoints should also receive domain-specific IDs.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/controller/AdminControllerV4.kt
```diff
@@ -238,6 +241,26 @@ class AdminControllerV4(
         return ResponseEntity.ok(TeamcitySyncJobResponse.from(state))
     }
 
+    /**
+     * Kick off a TeamCity Java/Maven validation run on the background executor.
+     * 202 with the freshly-started job, or (same-kind attach) the running job's
+     * state with 409; a cross-kind conflict is mapped to a structured 409 by
+     * [handleCrossKindConflict].
+     */
+    @PostMapping("/teamcity-validation")
```
pgorbachev, line 271: [P2] The runtime contract here is `202 Accepted` for a newly started job and `409 Conflict` when attaching to an existing job, but the generated `v4.json` documents a successful `200 OK` response.

The existing `POST /admin/teamcity-project-ids/sync` endpoint is documented the same way, so this appears to be a copied pre-existing pattern rather than a problem unique to this method. Since the specification is published for Portal code generation, please describe the actual responses explicitly for both endpoints with `@Operation`/`@ApiResponses` and regenerate the contract.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityProjectMapper.kt
```diff
@@ -0,0 +1,86 @@
...
+            buildType.template?.id?.let(::add)
+            buildType.templates?.buildTypes?.forEach { it.id.let(::add) }
+        }
+
+    private fun stepTypeOf(rawType: String): StepType =
+        when (rawType) {
+            "gradle-runner" -> StepType.GRADLE
+            "Maven2" -> StepType.MAVEN
+            "simpleRunner" -> StepType.COMMAND_LINE
+            else -> StepType.OTHER
```
pgorbachev, line 77: [P2] The validation module and SYS-071 support `StepType.IN_CONTAINER`, but this mapper never returns it: every raw runner type other than Gradle, Maven, and `simpleRunner` becomes `OTHER`.

Please map the actual TeamCity runner type and add a mapper test, or remove the unsupported enum path and requirement until the raw type is confirmed.

docs/registry/requirements-common.md
```diff
@@ -2367,7 +2377,245 @@ cases) and `ExternalTcProjectFetcherTest` (reference resolution, version-format,
...
+returns the `Set<ToolVersion>` (Java and/or Maven, for now) that step uses. Dispatch is by
+`StepType` (`DefaultBuildStepToolVersionResolver`): `MAVEN` → reads `maven.path` (Maven version)
+and `target.jdk.home` (Java version); `GRADLE` → reads only `target.jdk.home`; `COMMAND_LINE` and
+`IN_CONTAINER` (the latter an unconfirmed assumption — an in-container step is assumed to run a
+script the same way a command-line step does) → read `script.content`, split it on
+whitespace, and try *both* the Maven and the Java value-resolver on each (reference-resolved)
+token, since a bare command line gives no other signal about which tool a token belongs to.
+Every parameter read first goes through `ParameterReferenceResolver`, which recursively
+substitutes TeamCity `%paramName%` references within the same `Parameters` bag until no
+reference remains, returning `null` on a missing reference or a reference cycle (mirroring real
```
pgorbachev: [P2] This requirement says that a missing `%param%` reference returns `null`, but `ParameterReferenceResolver` deliberately preserves a missing reference as literal `%missing%` text, and its tests assert that behavior.

Please decide which behavior is the intended contract and align the requirement, implementation, and tests. The requirement cannot be marked tested while they disagree.

docs/registry/requirements-common.md
```diff
@@ -2367,7 +2377,245 @@ cases) and `ExternalTcProjectFetcherTest` (reference resolution, version-format,
...
+**Test method:** `ValidatorSuiteTest` — `SYS-068 throwing validator is isolated as an ERROR result`,
+`SYS-068 all-ok suite returns every result as-is`.
+
+### SYS-069: component-validation — JavaVersion.isEight and version normalization
+
+**Description:**
+`JavaVersion.isEight` is a literal test: `raw.trim() == "1.8" || raw.trim() == "8"` (per the
+implementation brief — no fuzzy matching at that layer). The burden of normalizing real-world
+values to that exact major-version form falls on `JavaVersionResolver` (decision **D1**): it
+extracts the major (or `1.<major>`) version from clean strings (`"11.0.2"` → `"11"`),
```
pgorbachev: [P2] `JavaVersionParser` does not exist—the implementation is `JavaVersionResolver`. More importantly, the resolver requires a `jdk`, `java`, or `jvm` marker, so bare values such as `11.0.2` and `8u392` return `null`, contrary to this requirement.

Please either implement and test the documented normalization shapes or narrow the requirement to the values actually supported.

component-validation/src/test/kotlin/org/octopusden/octopus/validation/teamcity/validator/MultipleJavaVersionValidatorTest.kt
```diff
@@ -0,0 +1,94 @@
...
+    private val validator = MultipleVersionValidator(
+        TeamCityValidationType.MULTIPLE_JAVA_VERSIONS,
+        configs,
+        steps,
+        DefaultBuildStepToolVersionResolver.standard(),
+        "Java",
+    ) { it is JavaVersion }
+
+    @Test
+    @DisplayName("NOT_APPLICABLE when there is nothing Java to inspect")
```
pgorbachev: [P2] These tests cover SYS-072, but neither the method names nor their `@DisplayName`s contain the requirement ID, which is inconsistent with the repository's test-to-requirement traceability convention. The documented "OK when no version resolves" case is also absent.

The same issues exist in `MultipleMavenVersionValidatorTest` for SYS-073. Please add the missing cases and include the requirement ID in both every test method name and `@DisplayName`.

docs/registry/requirements-common.md
```diff
@@ -70,7 +70,17 @@
...
+| SYS-071 | `component-validation` module: `BuildStepToolVersionResolver` — given one `BuildStep`, dispatch by `StepType` (Maven/Gradle/command-line+in-container) to read the right parameter(s), recursively resolve `%param%` references via `ParameterReferenceResolver`, and derive `Set<ToolVersion>` via the per-tool `ValueVersionResolver`s (`JavaVersionResolver`, `MavenVersionResolver`) | High | unit-test | ✅ Tested |
+| SYS-072 | `component-validation` module: `MULTIPLE_JAVA_VERSIONS` is WARNING when more than one distinct Java version is found across the same inspected steps as `USES_OLD_JAVA_VERSION`, OK for zero or one distinct version, NOT_APPLICABLE if nothing was inspectable | High | unit-test | ✅ Tested |
+| SYS-073 | `component-validation` module: `MULTIPLE_MAVEN_VERSIONS` is WARNING when more than one distinct Maven version is found across the same inspected steps, OK for zero or one distinct version, NOT_APPLICABLE if nothing was inspectable | High | unit-test | ✅ Tested |
```
pgorbachev: [P2] These requirements cover only the pure validation module. The server feature introduced by this PR—latest-only persistence, repeated replacement, stale pruning, post-sync triggering, lifecycle-gate behavior, job/event state, component-detail embedding, and dashboard APIs—has no numbered requirements or test mappings.

Please add server-side requirements and DB/controller integration tests, especially for consecutive replacement runs, final-project removal, duplicate version lines, and the 202/409 API contract.

components-registry-service-server/src/main/resources/db/migration/V7__add_teamcity_validation.sql
```diff
@@ -0,0 +1,12 @@
+-- Stored TeamCity validation findings, one row per (project_id, type). Only WARNING/ERROR
+-- results are persisted (OK / NOT_APPLICABLE are not stored); latest-only per project (each run
+-- replaces that project's rows). The (project_id, type) pair is the stable identity of a finding.
+-- See docs/teamcity-validation-server-integration-design.md.
```
pgorbachev: [P2] `docs/teamcity-validation-server-integration-design.md` does not exist in this repository. Please either add the design under `docs/registry/` and update this path, or remove the dangling reference.

The new table and server lifecycle should also be reflected in `technical-design.md`, `schema-spec.md`/ERD, the API changelog, and deployment/configuration documentation. Currently only validator requirements, OpenAPI, and TD-016 were updated.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationQueryService.kt
```diff
@@ -0,0 +1,78 @@
...
+            findings = rows.size,
+            byType = rows.groupBy { it.type }.mapValues { (_, v) -> distinctComponents(v) },
+            byStatus = rows.groupBy { it.status }.mapValues { (_, v) -> distinctComponents(v) },
+        )
+    }
+
+    /** project id -> the components (id + key) that own it, via version_line. */
+    private fun componentsByProject(projectIds: Set<String>): Map<String, List<Pair<UUID, String>>> =
+        versionLineRepository
+            .findByProjectIdsWithComponent(projectIds)
```
pgorbachev, line 84: [P3] The schema permits the same component/project pair to appear through multiple version lines. The normal sync path appears to avoid this in practice, but manual v4 curation can still create it, after which `list()` emits duplicate rows and the raw `findings` count is inflated.

A defensive `distinct` by `(projectId, componentId)` is inexpensive and would make this query correct for every schema-valid state. Please add a regression test for this edge case.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/EnrichedTcProjectFetcher.kt
```diff
@@ -0,0 +1,69 @@
...
+        val project: ExternalTeamcityProject?,
+        val at: Instant,
+    )
+
+    private val cache = ConcurrentHashMap<String, Cached>()
+
+    override fun fetch(projectId: String): ExternalTeamcityProject? {
+        val ttl = Duration.ofMinutes(validationProperties.cacheTtlMinutes)
+        val now = Instant.now()
+        cache[projectId]?.let { if (Duration.between(it.at, now) < ttl) return it.project }
```
pgorbachev, line 58: [P3] This cache survives across validation runs and is not invalidated when TeamCity sync completes. If a manual validation populates it and a sync runs within the next 30 minutes, the automatic post-sync validation can persist findings from the old TeamCity snapshot.

Each run already fetches each project ID only once, so the cache provides little benefit in the normal workflow. Consider scoping it to one run, invalidating it before post-sync validation, or bypassing it for that trigger.

docs/registry/requirements-common.md
```diff
@@ -70,7 +70,17 @@
...
+| SYS-074 | The component owner's manager (resolved via employee-service `getManager`) may edit the component and its field-overrides — a fourth, derived condition on `canEditComponent` alongside owner/RM/SC/admin. A directory failure or no-manager answer denies (fail-closed), never grants. `GET /{idOrName}/editors` enumerates the resolved manager (unlike the admin bypass, it is one concrete person per component); `getManager` is 2-minute cached per owner (resolved answers only) and its DB read runs in its own short-lived transaction, closed before the network call | High | unit + integration-test | ✅ Tested |
```
pgorbachev: [P3] SYS-064 was already published on `main` as the manager-edit requirement. Reassigning that ID to a different feature and moving the existing requirement to SYS-074 makes historical and external references ambiguous, even though the current Kotlin manager tests do not carry the requirement ID themselves.

Please preserve SYS-064 for the existing requirement and allocate subsequent IDs to the new validation requirements. The documented test-method references should then be aligned separately with the actual test names.

docs/registry/requirements-common.md
```diff
@@ -70,7 +70,17 @@
...
+| SYS-068 | `component-validation` module: `ValidatorSuite.validate` isolates a throwing validator as a single `Status.ERROR` result instead of losing every other check's result (decision D6) | Medium | unit-test | ✅ Tested |
+| SYS-069 | `component-validation` module: `JavaVersion.isEight` is true only for the exact `"1.8"` / `"8"` spellings; `JavaVersionParser` normalizes richer real-world values (`"1.8.0_392"`, `"JDK_ZULU_17_x64"`, `"/opt/java/openjdk-11"`, etc.) down to that major-version form before `isEight` is checked | Medium | unit-test | ✅ Tested |
+| SYS-070 | `component-validation` module: `TeamCityValidators` (the suite) returns exactly the five TeamCity results for a given project, each with the status the individual checks would produce in isolation | High | unit-test | ✅ Tested |
```
pgorbachev: [P3] This summary says the suite returns five TeamCity results, while the detailed SYS-070 section and `TeamCityValidators` correctly define six checks. Please update the summary to six so the requirement is internally consistent.

pgorbachev (changes requested):
The validation module boundary and one-way dependency are well designed, but the current server integration contains a blocking repeated-run persistence bug and several lifecycle, API-contract, configuration, and documentation gaps. Details are provided inline. Please address the P1 issue and the substantive P2 findings before merge.

components-registry-service-server/src/main/resources/application.yml
```diff
@@ -100,6 +100,15 @@ teamcity:
...
   sync:
     cron: "0 0 4 * * SUN"
+  validation:
+    gradle-build-template-id: CDGradleBuild
+    maven-build-template-id: CDJavaMavenBuild
+    release-family-template-ids:
+      - CDRelease
+      - CdReleaseCandidateNew
+      - CdReleaseChecklistValidation
+    gradle-default-build-step-id: GRADLE_ID
```
pgorbachev: [P2] This entire `teamcity.validation` block describes environment-specific TeamCity topology and should not be hardcoded in the service sources. Besides the placeholder default-step IDs, the template IDs and release-family IDs (`CDRelease`, `CdReleaseCandidateNew`, `CdReleaseChecklistValidation`) are internal system details that may vary by environment. Bundling them makes a deployment without external configuration silently use repository assumptions; with `GRADLE_ID` / `MAVEN_ID`, `DefaultBuildStepResolver` returns `null`, `OVERRIDES_DEFAULT_BUILD_STEP` can become `OK`, and the default-step contribution to Java-version checks is skipped. Please move all `teamcity.validation.*` values to `service-config`. The source-level baseline should leave validation explicitly disabled or unconfigured, and startup/activation should validate that every required value is present and is not a placeholder. Test profiles can supply fixture values. The upgrade note should document the real per-environment values and activation procedure.

components-registry-service-server/src/main/resources/openapi/v4.json
```diff
@@ -5569,6 +5745,435 @@
...
+          },
+          "409" : {
+            "content" : {
+              "*/*" : {
+                "schema" : {
+                  "$ref" : "#/components/schemas/ErrorResponse"
+                }
+              }
+            },
+            "description" : "A TC validation (or a conflicting migration kind) is already running; body describes the in-flight job/conflict"
```
pgorbachev: [P2] The status codes are now correct, but the documented `409` response schema is still inaccurate. A same-kind retry returns `TeamcityValidationJobResponse`, while a cross-kind conflict returns `MigrationConflictResponse`; the committed OpenAPI currently advertises `ErrorResponse` for both. Generated clients will therefore model the attach response incorrectly. Please describe the `409` content using the two actual response shapes, for example with `oneOf`, and apply the equivalent correction to the sibling TeamCity sync endpoint.

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationRepeatedRunIntegrationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    fun stale_row_pruning_with_empty_scope_removes_the_last_project() {
+        val projectA = "FINAL-REMOVAL-PROJECT-A"
+        val projectB = "FINAL-REMOVAL-PROJECT-B"
+        replace(projectA, listOf("USES_OLD_JAVA_VERSION"))
+        replace(projectB, listOf("OVERRIDES_DEFAULT_BUILD_STEP"))
+        assertEquals(2, teamcityValidationRepository.findByProjectIdIn(listOf(projectA, projectB)).size)
+
+        // Scope shrinks to empty (both projects unlinked) — mirrors
+        // TeamcityValidationService.removeStaleProjects(knownProjectIds = emptySet()).
+        val knownProjectIds = emptySet<String>()
```
pgorbachev: [P2] This test reimplements the stale-set subtraction and calls the repository directly instead of exercising `TeamcityValidationService`. It would still pass if the original `knownProjectIds.isEmpty()` early return were reintroduced, so it does not actually guard the SYS-087 regression. Please drive the real service with an empty linked-project scope and assert that all stored findings are removed. Also include `SYS-087` in the method name and `@DisplayName` to follow the repository's test-to-requirement traceability convention.

docs/registry/api-changelog.md
```diff
@@ -18,6 +18,15 @@ refresh it with `./gradlew :components-registry-service-server:generateOpenApiDo
 
 ## Unreleased
 
+- **TeamCity validation surfaced in the API — coordinated deploy required.** `TeamcityProjectResponse`
+  gains `validations: List<ValidationResponse>` (`type` + `status` + optional `message` +
+  optional `updatedAt`, e.g. `USES_OLD_JAVA_VERSION` / `WARNING`), and new admin endpoints
+  `GET /rest/api/4/admin/teamcity-validations` and `.../summary` expose per-project findings
+  (`@permissionEvaluator.canImport()`). Findings are populated by a new post-sync + on-demand
+  validation job — see [Upgrade note](deployment/tc-validation-upgrade-note.md) for the required
```
pgorbachev, line 26: [P2] This coordinated deployment depends on correctly configured `teamcity.validation.*` values, but the referenced `docs/registry/deployment/tc-validation-upgrade-note.md` is not included in the PR. The technical design also links to the nonexistent `tech-debt/016-validator-token-matching.md`; the actual validator follow-up is `016-component-validation-version-resolver-value-coverage.md`, while TD-016 is now also assigned to the ArchUnit package-cycle debt merged from main. Please add the upgrade note with the real TeamCity IDs and rollout procedure, allocate a unique TD number, and update all references.

docs/registry/requirements-common.md
```diff
@@ -8,70 +8,88 @@
...
+| SYS-086 | TC validation scope (which projects to validate and to prune stale rows for) is derived from `version_line` — the projects currently linked to a component — not from the append-only `teamcity_project` table, which never removes a project once unlinked                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | High | unit-test | ❌ Not tested |
```
pgorbachev: [P2] The server-side requirements are now documented, but SYS-086, SYS-088, SYS-089, and SYS-091 are explicitly left untested. These cover the linked-project scope, post-sync trigger and cache invalidation, lifecycle-gate exclusion, and component-detail embedding—the main integration seams of the feature. Please add focused coverage before treating the server integration as complete. The tests recorded for SYS-085, SYS-090, and SYS-092 should also carry their requirement IDs in their method names and `@DisplayName` values to follow the repository's traceability convention.

docs/registry/requirements-common.md
```diff
@@ -2368,6 +2386,256 @@ cases) and `ExternalTcProjectFetcherTest` (reference resolution, version-format,
 build-type fallback, archived/all-paused exclusion); v4 PATCH preservation in
 `ComponentManagementServiceImpl` write-path tests.
 
+### SYS-075: component-validation — ATTACHED_TO_BUILD_TEMPLATE
+
+New pure-Kotlin module `component-validation` validates
+TeamCity projects: input in (`TeamcityProject`), `List<ValidationResult>` out — no Spring, no
+DB, no TeamCity-client dependency, no IO. `ATTACHED_TO_BUILD_TEMPLATE` is the first of its five
```
pgorbachev: [P3] SYS-081 and `TeamCityValidators` now correctly describe six checks, but this detailed SYS-075 section still calls `ATTACHED_TO_BUILD_TEMPLATE` the first of five checks. Please change "five" to "six".

pgorbachev (changes requested):
Thanks for the thorough update. The original 17 findings were addressed in good faith: 14 are now resolved, and the three partially addressed threads are represented by focused follow-ups below. The bulk-delete/repeated-run fix, linked-project scope correction, DTO and operation-ID changes, cache invalidation, unsupported runner cleanup, and requirement renumbering all look good. All checks are green on this head. I am requesting changes for the remaining false-clean configuration path, inaccurate 409 OpenAPI contract, and the stale-pruning regression test that does not exercise the service; the other comments are documentation and coverage cleanup.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/teamcity/TeamcityValidationProperties.kt
```diff
@@ -0,0 +1,41 @@
...
+@ConfigurationProperties(prefix = "teamcity.validation")
+class TeamcityValidationProperties(
+    /** Build template id whose configs are "our Gradle build". */
+    @field:NotBlank
+    val gradleBuildTemplateId: String,
+    /** Build template id whose configs are "our Maven build". */
+    @field:NotBlank
+    val mavenBuildTemplateId: String,
+    /** Release-family templates excluded from "not attached to build template". */
+    @field:NotEmpty
```
pgorbachev, line 39: **[P2] Reject blank release-family IDs**

`@NotEmpty` only verifies that the collection contains at least one element; it does not validate the elements themselves. For example, `release-family-template-ids: [""]` would pass startup validation but match no real release-family template. Release configurations would then appear in `notAttachedToBuildTemplate`, producing false WARNING noise that the fail-fast contract is intended to prevent. Please add container-element validation for non-blank IDs (for example, the equivalent of `Set<@NotBlank String>`) and cover both blank and whitespace-only elements in `TeamcityValidationPropertiesTest`.

docs/registry/deployment/tc-validation-upgrade-note.md
```diff
@@ -0,0 +1,67 @@
...
+boot.
+
+Note: declaring the fields as non-null Kotlin `String`s does NOT enforce this on its own — a blank
+YAML value like `gradle-build-template-id:` binds to `""` (Spring converts a null YAML scalar to an
+empty string), which is non-null and would bind fine. The `@NotBlank`/`@NotEmpty` constraints are
+what actually reject blanks at startup.
+
+## Per-environment values (service-config)
+
+The repo `application.yml` carries the current TeamCity defaults so local/dev/test contexts
```
pgorbachev: **[P2] Align the upgrade note with the current configuration baseline**

This section no longer matches the implementation. The bundled `application.yml` does not carry working TeamCity defaults or `GRADLE_ID` / `MAVEN_ID` placeholders anymore: all five values are intentionally blank/empty, and production startup depends on external `service-config`. The statement that `test-db-validate` inherits those defaults is also stale because test contexts now receive explicit fixture values from their dedicated test configuration. Please update the note to describe the actual fail-fast baseline, the required service-config-before-deployment ordering, and the current test-profile setup.

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/teamcity/validation/TeamcityValidationQueryServiceTest.kt
```diff
@@ -0,0 +1,116 @@
...
+        assertEquals(5, rows.size)
+        assertEquals(setOf(a, b), rows.map { it.componentId }.toSet())
+        assertEquals(1, rows.count { it.projectId == "Bar" }) // Bar owned only by comp-a
+        assertEquals(4, rows.count { it.projectId == "Foo" }) // 2 findings x 2 components
+        assertEquals("https://tc.example.com/project/Foo", rows.first { it.projectId == "Foo" }.projectUrl)
+        assertEquals("https://tc.example.com/project/Bar", rows.first { it.projectId == "Bar" }.projectUrl)
+    }
+
+    @Test
+    @DisplayName("list filters by validation type before the component join")
```
pgorbachev: **[P2] Complete the declared SYS-092 filter coverage**

SYS-092 acceptance criterion 3 declares verification of the `type`, `status`, and `component` filters, but this test class currently exercises only `type`. Please add explicit cases for `status` and `component` before keeping the requirement at `✅ Tested`. Since these methods are the requirement's direct query-semantics coverage, please also include `SYS-092` in their method names and `@DisplayName` values as required by the repository's test-to-requirement traceability convention.

components-registry-service-server/src/test/kotlin/org/octopusden/octopus/components/registry/server/service/impl/MigrationLifecycleGateTest.kt
```diff
@@ -89,6 +89,41 @@ class MigrationLifecycleGateTest {
         assertEquals(JobKind.HISTORY, gate.current()!!.kind)
     }
 
+    @Test
+    fun `SYS-089 same-kind TC_VALIDATION second tryClaim attaches to the existing owner`() {
```
pgorbachev, line 95: **[P3] Add the missing SYS-089 display names**

These new tests include `SYS-089` in their Kotlin method names but do not provide the corresponding `@DisplayName("SYS-089: ...")`. Please add the display names so the new coverage follows the repository's documented test-to-requirement traceability convention and remains identifiable in generated test reports.

pgorbachev (changes requested):
The feature is now structurally sound, and the six findings from the previous round have been addressed correctly: runtime-specific TeamCity IDs were moved out of the bundled configuration, the 409 OpenAPI contract was corrected, SYS-087 now exercises the real service path, the missing integration seams received coverage, the documentation files and references were restored, and the five-versus-six inconsistency was fixed.

The remaining findings are finishing issues rather than architectural defects. There is no P1 runtime blocker, and CI is fully green. However, the mandatory configuration contract still accepts a blank release-family entry, and the deployment note currently gives instructions that contradict the final fail-fast implementation. SYS-092 is also marked fully tested without covering two of its declared filters, while SYS-089/SYS-092 traceability remains incomplete.

I recommend addressing the P2 configuration, deployment-documentation, and SYS-092 coverage issues before approval. The P3 traceability comment should preferably be completed in the same round.

components-registry-service-server/src/main/kotlin/org/octopusden/octopus/components/registry/server/config/ApplicationConfig.kt
pgorbachev, line 1: **[P2] Do not require validation configuration in no-DB mode**

`TeamcityValidationProperties` is registered unconditionally, while the bundled values are intentionally blank and validated with `@NotBlank`/`@NotEmpty`. As a result, a pure Git/no-DB deployment can fail at startup unless it supplies configuration for a feature it cannot use. The current `NoDbModeContextTest` masks this because the `common` profile provides fixture values. Please register the properties and helper beans only when the database is enabled, and add a no-DB context test that boots without any `teamcity.validation.*` overrides. Please also update the upgrade note to clarify that Git/no-DB instances do not need these values.

component-validation/src/main/kotlin/org/octopusden/octopus/validation/validators/teamcity/RelevantBuildSteps.kt
pgorbachev, line 1: **[P3] Document the message format as a cross-repository contract**

The KDoc still describes “step name (step id) in configuration name (configuration id)”, while the implementation emits `- STEP_ID in CONFIGURATION_ID`. Portal parses this exact shape to generate TeamCity links, so it is now a cross-repository wire contract. Please update the KDoc and explicitly note that changing this format requires a coordinated Portal update.

docs/registry/deployment/tc-validation-upgrade-note.md
pgorbachev, line 1: **[P3] Align the upgrade note with the actual validation mechanism**

The note says blank collection elements are rejected through `Set<@NotBlank String>`, but the implementation deliberately uses an `@AssertTrue` bean-property check because Kotlin container-element annotations were not reliably handled by Hibernate Validator. The documented behavior is correct, but the mechanism should match the code.

pgorbachev (changes requested):
The validation architecture remains strong: the pure validation module is separated cleanly from server-side IO, the system-specific template IDs are externalized to `service-config`, and the previous review findings are addressed. One operational issue remains: mandatory validation properties are currently enforced even in no-DB/Git-only mode, where the feature is unavailable. The other comments are documentation corrections.

pgorbachev (dismissed):
Approved after the follow-up review. The latest findings are addressed, the architecture and runtime behavior look sound, and the Merge Gate is green. Release and deployment should remain coordinated with [Portal #200](https://github.com/octopusden/octopus-components-management-portal/pull/200).

pgorbachev (approved):
Approved after re-reviewing the update from main. The conflict resolution preserves both the existing SYS-067 changes and the TeamCity validation requirements/fixes, the follow-up commit only realigns the ktlint baseline line numbers, and the full Merge Gate is green. Portal #200 should still be completed and released in coordination with this backend change.

PR #455, opened 2026-07-24
Add post Github commit status step

gnh374:
## Summary by CodeRabbit

* **New Features**
  * Added an automated “[2.9] Step 2 Aggregate [AUTO]” build configuration.
  * It now triggers on successful completion of the compile step across all branches, aggregates results from validation, integration DB tests, and compatibility suites, and snapshots required builds.
  * Build numbering is propagated from the upstream compile build, and dependency failures are reported as problems for clearer tracking.
  * Configured to use the standard GitHub status template for consistent reporting.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -1431,10 +1431,6 @@ object id29Step2AggregateAuto : BuildType({
             reuseBuilds = ReuseBuilds.SUCCESSFUL
             onDependencyFailure = FailureAction.ADD_PROBLEM
         }
-        snapshot(id30DeployToOkdQaDevAuto) {
-            reuseBuilds = ReuseBuilds.SUCCESSFUL
-            onDependencyFailure = FailureAction.ADD_PROBLEM
-        }
     }
 })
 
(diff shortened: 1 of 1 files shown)


.teamcity/settings.kts
pgorbachev, line 1: 1. Please declare `id29Step2AggregateAuto` as `BuildTypeSettings.Type.COMPOSITE`. It has no build steps and only aggregates snapshot dependency results, so keeping the default `REGULAR` type unnecessarily requires a build agent and checkout for every run.

2. Please remove `snapshot(id30DeployToOkdQaDevAuto)` from this aggregate. Since `id29` runs for every branch via `branchFilter = "+:*"`, keeping this dependency would also start the QA DEV deployment for feature branches. The deployment should retain its previous behavior and run only on the default branch through its own finish trigger.

pgorbachev (changes requested):
Please address the review comments before merging.

.teamcity/settings.kts
gnh374: the id29Step2AggregateAuto is attached to the RDDepartment_PostGithubStatus template, which posts the commit status via a PostGithubCommitStatus build step — and a build step only runs on an agent. A composite build type runs on no agent and executes no build steps, so switching id29 to composite would silently skip the status post and  the aggregate would never report anything back to GitHu

pgorbachev (approved):

rodemark (approved):

PR #456, opened 2026-07-25
chore(publish): stop publishing deployables and tooling to Maven Central

pgorbachev:
## Why

The org is over Maven Central's **monthly** publishing limits — Release Size 1.21 GB vs 80 MB, File Count 10 094 vs 1 000, Release Count 32 vs 7. Measured directly against Central, **CRS is the single biggest contributor**: 670 files and 39.8 MB per release ⇒ ~6 000 files and ~358 MB/month at 9 releases (July).

Most of that is artifacts **nobody consumes as a Maven dependency**. Central is an exchange for dependencies; deployables and internal tooling don't belong there.

## Measured footprint (3.0.8, from repo1)

| module | files | size | |
|---|--:|--:|---|
| components-registry-automation | 70 | 19.75 MB | **keep** (see below) |
| **components-registry-cli** | 60 | **16.43 MB** | **drop** |
| **components-registry-service-server** | 60 | **2.00 MB** | **drop** |
| component-resolver-core | 50 | 0.62 MB | keep |
| components-registry-api | 50 | 0.30 MB | keep |
| component-resolver-api | 50 | 0.27 MB | keep |
| components-registry-service-light-client | 60 | 0.17 MB | keep |
| components-registry-service-core | 50 | 0.11 MB | keep |
| components-registry-dsl | 50 | 0.08 MB | keep |
| test-common | 60 | 0.05 MB | keep |
| components-registry-service-client | 60 | 0.04 MB | keep |
| **components-registry-compat-test** | 50 | 0.01 MB | **drop** |
| **total** | **670** | **39.83 MB/release** | |

## What is dropped, and why it's safe

- **components-registry-service-server** — the deliverable is the docker image on ghcr. Consumers needing `openapi/v4.json` read it **from git at a pinned ref** (portal `frontend/scripts/vendor-spec.sh`), not from the artifact. `.teamcity/settings.kts` already notes the server's Maven publication is unusable for the compat baseline (non-boot jar) and uses `docker pull` + `docker cp` instead.
- **components-registry-cli (crsctl)** — no consumers anywhere in the org. Its README documents building locally (`:components-registry-cli:shadowJar`) and defers release/packaging to a separate follow-up. Shadow + Gradle module metadata were dragging a **15.9 MB `-all` fat jar** into Central through the `shadowRuntimeElements` variant of `components.java`.
- **components-registry-compat-test** — internal compatibility harness, never a dependency.

**`component-validation`** (added to `main` by #443 after this branch was opened) is also **not** published: it has never been on Central, the server consumes it as `project(':component-validation')` which needs no publication, and nothing in the organisation resolves it as a dependency. Excluding it declines to *start* publishing rather than dropping an artifact; the allowlist comment names the trigger to revisit — the first consumer outside this repository.

The 8 consumed library modules still publish and sign **exactly as before** (verified: the diff is re-indentation only for them).

## Kept on purpose: components-registry-automation

Even though it's the heaviest artifact (19.75 MB), it **must stay**: its TeamCity metarunner resolves the fat jar from a Maven repo at build time —

```xml
-Dartifact=${group}:${name}:${version}:jar:all
java -jar ./.$name/$name-$version-all.jar %ARGS%
```

(`components-registry-automation/metarunners/OctopusComponentsRegistryAutomation.xml`). A plain dependency search doesn't reveal this — consumption goes through the metarunner.

## Implementation note

Declaring **no publication** is what actually keeps a module out of Central. The module-level `tasks.named('publish') { enabled = false }` that compat-test already had was **ineffective** — `publishToSonatype` depends on the `PublishToMavenRepository` tasks, not on the per-project `publish` lifecycle task — which is exactly why compat-test was on Central despite the intent. Those two dead lines are removed.

## Effect

| | before | after |
|---|--:|--:|
| files / release | 670 | **500** |
| MB / release | 39.83 | **21.39** |
| at 9 releases/month | 6 030 files, 358 MB | **4 500 files, 193 MB** |
| **saved** | | **−1 530 files, −166 MB / month** |

## Verification

- Full build green: `AUTH_SERVER_URL=… ./gradlew build -x dockerBuildImage -x dockerPushImage -x ocCreate -x ocProcess -x :components-registry-automation:test` → `BUILD SUCCESSFUL`
- `publishToMavenLocal --dry-run` → exactly **9** `publishMavenJavaPublication*` + **9** `signMavenJavaPublication` tasks; the 3 dropped modules have neither
- Independent review (Sonnet): no blockers; Groovy scoping, `publish`-task wiring at build.gradle:553-560, `.teamcity/settings.kts`, and mavenLocal resolution all checked

## Deliberately out of scope

`withJavadocJar()/withSourcesJar()` still run for the 3 unpublished modules — wasted work, but TeamCity artifact rules may expect those jars and the saving is seconds. Left as-is.

## Follow-up

After this, **92% of CRS's remaining Central size is automation's 19.7 MB fat jar**, which lives there only because TeamCity fetches it from a Maven repo. If TC agents resolve through the internal Artifactory (the metarunner uses `userSettingsSelection:default`), publishing that fat jar internally instead would save a further ~178 MB/month. Needs agent-settings verification — separate task.

## Summary by CodeRabbit

* **Chores**
  * Updated release configuration to publish/sign only an allowlisted set of modules to Maven Central.
  * Added a build-time safeguard that fails the build if the actual publishing set drifts from the allowlist.
  * Preserved required thin/thick JAR outputs for CI and release tooling.
  * Adjusted the compatibility-test module to no longer explicitly disable publishing tasks.
  * Refreshed CLI documentation to reflect current build/download status and clarify fat-jar naming/classifiers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,19 +23,24 @@ plugins {
     id 'com.platformlib.gradle-wrapper' version '0.2.6'
 }
 
-// Modules deliberately NOT published to Maven Central. Central is an exchange for artifacts
-// other projects consume as a Maven dependency; these are deployables and internal tooling:
-//   - components-registry-service-server: the service itself. Its deliverable is the docker
-//     image on ghcr; the published jar is a 1.5 MB deploy artifact nobody depends on. (Consumers
-//     reference its openapi/v4.json by git path, not via the artifact.)
-//   - components-registry-cli (crsctl): standalone CLI. README documents building it locally
-//     (`:components-registry-cli:shadowJar`); packaging/distribution is a separate follow-up.
-//     Shadow + Gradle module metadata were dragging its 15.9 MB `-all` fat jar into Central.
-//   - components-registry-compat-test: internal compatibility harness, never a dependency.
-// NOT in this list: components-registry-automation — its TeamCity metarunner resolves
-// `<group>:components-registry-automation:<version>:jar:all` from a Maven repo at build time,
-// so it must stay published (see components-registry-automation/metarunners/).
-def unpublishedModules = ['components-registry-service-server', 'components-registry-cli', 'components-registry-compat-test']
+// Modules published to Maven Central — an allowlist, so a module that is not listed is not
+// published. Central is for artifacts other projects consume as a Maven dependency; NOT listed
+// here on purpose: components-registry-service-server and components-registry-cli (deployables,
+// shipped as a docker image / built locally per the CLI README) and components-registry-compat-test
+// (internal harness). components-registry-automation IS listed even though it is tooling: its
+// TeamCity metarunner resolves `<group>:components-registry-automation:<version>:jar:all` from a
+// Maven repo at build time (see components-registry-automation/metarunners/).
+def centralPublishedModules = [
+        'component-resolver-api',
+        'component-resolver-core',
+        'components-registry-api',
+        'components-registry-dsl',
+        'components-registry-service-client',
+        'components-registry-service-light-client',
+        'components-registry-service-core',
+        'test-common',
+        'components-registry-automation',
+] as Set
 
 octopusQuality {
     // NOTE: in octopus-quality 2.5.2 `excludeProjects` maps ONLY to coverage exclusion
@@ -302,12 +307,10 @@ configure(subprojects) {
         withSourcesJar()
     }
 
-    // Maven Central is for modules other projects consume as a DEPENDENCY. Deployables and
-    // internal tooling are deliberately not published there (see unpublishedModules above):
-    // creating no publication is what keeps them out — `publishToSonatype` then has nothing
-    // to upload for them, while `build`, the docker image and the shadow/boot jars are
-    // untouched.
-    if (!(project.name in unpublishedModules)) {
+    // Only allowlisted modules get a publication (see centralPublishedModules above); for the
+    // rest `publishToSonatype` has nothing to upload, while `build`, the docker image and the
+    // shadow/boot jars are unaffected.
+    if (project.name in centralPublishedModules) {
         publishing {
             publications {
                 mavenJava(MavenPublication) {
@@ -436,6 +439,21 @@ ${sectionsHtml}
 </body>
 </html>
 """
+}
+
+// Fail the build if any module publishes to Maven Central outside the allowlist. A
+// publication-surface reduction has no test of its own: without this guard, dropping the
+// allowlist check or adding a module-level `publishing {}` block silently starts pushing a
+// deployable to Central again, and it is only noticed at the next quota audit.
+gradle.projectsEvaluated {
+    def actual = subprojects.findAll { !it.publishing.publications.isEmpty() }*.name as Set
+    if (actual != centralPublishedModules) {
+        throw new GradleException(
+                "Maven Central publication set drifted.\n" +
+                        "  allowlisted: ${centralPublishedModules.sort()}\n" +
+                        "  publishing:  ${actual.sort()}\n" +
+                        "Update `centralPublishedModules` in the root build.gradle only if the change is intentional.")
+    }
 }
 
 // Aggregate JaCoCo tasks and coverage gate wired in projectsEvaluated (outside any task
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,20 @@ plugins {
     id 'com.platformlib.gradle-wrapper' version '0.2.6'
 }
 
+// Modules deliberately NOT published to Maven Central. Central is an exchange for artifacts
+// other projects consume as a Maven dependency; these are deployables and internal tooling:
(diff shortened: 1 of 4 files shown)


pgorbachev:
Pushed a second commit addressing the review findings.

**Codified the invariant (the review's test-coverage gap).** The module list is now an **allowlist** (`centralPublishedModules`) rather than a denylist, plus a `projectsEvaluated` guard that fails the build if the set of modules declaring a `MavenPublication` differs from it.

Why the allowlist and not just a guard over the denylist: with a denylist the same list feeds *both* the publication creation and the expectation, so the two can never disagree — my first attempt at the guard was tautological and a RED demo proved it (editing the list did not fail the build). With an allowlist a newly added module defaults to **not** published, which is the safe direction, and the guard catches any publication appearing outside the list.

Verified both ways:

```
GREEN  ./gradlew help                       → exit 0, guard silent
RED    condition forced to `if (true)`      → build fails:
       Maven Central publication set drifted.
         allowlisted: [9 modules]
         publishing:  [... + components-registry-cli, components-registry-compat-test,
                       components-registry-service-server]
```

**Other findings fixed:**
- `components-registry-cli/README.md` — dropped the now-false "and Maven publication" clause (thanks, that was missed when the matching build-script comment was fixed), and added a "Prebuilt download" row: the module is not on Central and is built locally until a distribution channel exists. Dropping a runnable tool from Central removes an install path, so the replacement (a GitHub Release asset — the release pipeline already creates the GitHub release) shouldn't stay implicit. Deliberately future, not promised in this PR.
- `components-registry-cli/build.gradle` — comment now references the renamed variable.
- Root comment trimmed to what the list means; the MB statistics live in this PR description, per the repo convention for config comments.
- Guard simplified per the nit: the `hasPlugin('maven-publish')` check was dead, since that plugin is applied to every subproject.

Full build green after each step (`BUILD SUCCESSFUL`), and the second review pass found no blockers.

pgorbachev:
Fixed the P3: dropped the `~16 MB of the org's monthly Central quota` figure from `components-registry-cli/README.md:19`. You were right that it was still there on `1315d494` — CodeRabbit had marked it resolved, but the number I removed earlier was in the *build-script* comment, not this README row. Same convention applies to both: files carry meaning, statistics live in the PR description. The reason the CLI is not on Central — a Maven coordinate is the wrong channel for a CLI — stays.

Thanks for the independent confirmation on the mechanics: `publishToSonatype --dry-run -Pnexus` / `publishToMavenLocal --dry-run` green, publish+sign tasks created only for the 9 allowlisted modules, empty lifecycle tasks for the excluded three, and no consumers of the dropped coordinates anywhere in the org.

adellinekania (dismissed):

pgorbachev:
Both blocking findings addressed — thanks for holding the approval, each would have shipped a real problem.

## 1. Stale base / `component-validation`

Rebased onto current `main` (was `57043ba3`, missing #443, #463 and #466). You were right that the module loses its publication after the merge — my own RED demo now shows it explicitly in the drift list, which is only visible *because* of the rebase.

Decided explicitly and documented next to the allowlist: **it stays unpublished.** The facts that settle it:

- it has **never been on Central** — `org/octopusden/octopus/infrastructure/component-validation/` is `404`, since it landed after the last release;
- the server consumes it as **`project(':component-validation')`** (`components-registry-service-server/build.gradle:17`), which needs no publication at all;
- no repository in the organisation resolves it as a Maven dependency.

So excluding it is *declining to start* publishing, not dropping an existing artifact — and publishing it would cost ~50 files per release (≈450/month at the current cadence) for something nobody can resolve. Given #443's "built to serve other services later", the comment names the trigger to revisit: the first consumer outside this repository, at which point it is a one-line addition.

## 2. Guard moved out of configuration time

Fully agreed, and worse than a style issue: #466 had just removed this exact pattern from this exact file, leaving the rule in a comment — *a policy problem must fail its gate, not every Gradle invocation*. My guard would have reintroduced it into freshly repaired code.

Reworked to the same shape #466 uses: problems collected in a closure, surfaced through a **`verifyCentralPublicationPolicy`** task, wired into the publish path (`publishToSonatype` / `publish` / `publishToMavenLocal`, matched by name so neither task is forced into existence). Drift still cannot reach Central, since the gate runs before any upload.

Verified in both directions:

```
policy holds                        → verifyCentralPublicationPolicy exit 0
allowlist check forced off:
  ./gradlew help                    → exit 0        ← the point of the change
  verifyCentralPublicationPolicy    → exit 1, listing component-validation,
                                      cli, compat-test, server as unexpected
gate present in publishToMavenLocal task graph → yes
full build                          → BUILD SUCCESSFUL
publication count                   → still exactly 9
```

The PR description is updated for both: the allowlist rationale now covers `component-validation`, and the guard is described as a task rather than a configuration-time check.

fishinitself (approved):

PR #464, opened 2026-07-27
ci(mutation): add PIT mutation testing — per-PR check on GitHub, weekly trend on TeamCity

pgorbachev:
Adds PIT mutation testing over the server's pure-logic packages: a per-PR **check** on GitHub (reporting, not blocking) and a weekly **analytics** build on TeamCity for the report and the score trend.

Base is `main`; the coverage-floor work (#463, #466, #467) has merged and this branch is rebased on top,
so the diff below is only the mutation-testing change.

## Why

Coverage counters answer "did this line run". They cannot answer "would any test have noticed it behaving differently" — and this module reports 89.5% LINE / 70.7% BRANCH, so there is room for code that is executed but unchecked. The first run found exactly that: of 1618 mutants, **342 SURVIVED** — the tests reach the code and nothing checks the result. Dominant mutators among them are **VoidMethodCall 155** and **NegateConditionals 127**: a call can be dropped or a condition inverted and the suite stays green.

The starkest case is on the v1/v2/v3 wire path: `JiraComponentVersionToDetailedComponentVersionMapper.convert` has **23 of 27** mutants surviving, all of the "removed call to setX" shape — every field assignment in that mapper can be deleted unnoticed by unit tests.

## What

**Gradle** — `info.solidsoft.pitest` in `pluginManagement`, versions in `gradle.properties`, applied in the server module only. Scope is `..server.util..` and `..server.mapper..`: PIT re-runs the covering tests once per mutant, so it is affordable only over pure logic, and meaningful only where a wrong result is a wrong *answer* rather than wrong wiring. Controllers, services, repositories, config and entities are not targeted.

Thresholds are **floors**, not ratchets: `mutationThreshold = 53`, `coverageThreshold = 78`, both set below the measured values. Nothing in the build prevents lowering them, so holding the numbers is a review property — the threshold sits in the diff. 53 is a floor to raise, and the report separates the two kinds of work needed: killing SURVIVED mutants strengthens assertions, while NO_COVERAGE means unit tests never reach that code (the integration-tagged suite that does cover it is excluded here).

**Two settings are load-bearing:**

- `excludedGroups = ['integration', 'performance']` — those tests live in the **same `src/test` source set** and are filtered by *tag* in the Gradle tasks. PIT drives the JUnit platform itself and inherits none of that, so without this it boots Testcontainers and a Spring context **once per mutant**.
- `excludedClasses = ['*$$inlined$*']` — Kotlin emits one comparator/lambda holder per `sortedBy` call site; those have no source line to write a test against.

**GitHub** — `.github/workflows/mutation.yml`. A **check, not a gate**: it reports on the PR and is deliberately absent from `gate-merge.needs`, so a red run is visible and does not block a merge. Manual, weekly, and on PRs touching the targeted sources, the server's tests or the build files. `permissions: contents: read`, since the job runs PR-controlled Gradle logic. A concurrency group cancels superseded runs on the same ref. Publishes the HTML report and a killed/survived/not-covered table in the job summary, counting `detected='true'` — what `mutationThreshold` is applied to, not `status='KILLED'`.

**TeamCity** — `[1.1] Mutation Testing [AUTO]`, **outside the build chain**: no snapshot dependency in either direction, no `finishBuildTrigger`, and none of the project's snapshot dependencies points at it, so it can neither delay nor block `[2.4]` deploy or `[4.0]` release. Weekly on the default branch, Sunday 11:00 UTC; the inherited per-commit and all-branch triggers are disabled. `[AUTO]` because in this project the suffix says whether a config fires by itself.

An empty `dependencies {}` block is deliberately **not** used to "pin" the off-chain property: TeamCity does not let a build type drop an inherited setting, so such a block would generate nothing and guarantee nothing. What holds instead is documented next to the claim — the shared template declares no dependencies today, and were it to gain one, every config using it inherits it, making that a project-wide visible change rather than a silent capture of this build.

The two homes split by purpose, not by duplication:

| | GitHub | TeamCity |
|---|---|---|
| Role | per-change **signal** on the diff | **analytics**: report and trend |
| When | every PR touching the paths | weekly on the default branch |
| Gives | check result + job summary | browsable HTML in the artifact tree + `mutationScore` charted per build |

TeamCity also covers a blind spot the GitHub trigger structurally has: that workflow fires only on changes to the server module and the build files, so a change elsewhere (`component-resolver-core`, which the mappers call) can move the score without ever running it. Widening the paths would put an 8-minute analysis on every PR, including doc-only ones; a weekly backstop is far cheaper.

A release gate on the score is deliberately absent and should stay absent: releases are cut from the default branch, every commit there passed the PR checks, and test quality is fixed in review rather than on the release path.

## Verification

- `:components-registry-service-server:pitest` green: 1618 mutants, 895 detected (55%), both floors pass. Full `build qualityStatic :…:pitest` green.
- **Calibration was measured, not guessed:** the first run used placeholder thresholds and failed on `coverageThreshold` (`Line coverage of 80 is below threshold of 85`) — that is how the real values were found.
- **The check fails loudly if its scope disappears:** pointing `targetClasses` at renamed packages gives `Created 0 mutation test units` → `PitHelpError: No mutations found` → `BUILD FAILED`, rather than a green run over an empty set.
- **PIT's XML quoting was verified against the pinned engine**, not assumed: the compiled `XMLReportListener` in pitest 1.16.1 emits single-quoted attributes, which is what both the workflow's grep and the TeamCity stats script rely on.
- TeamCity DSL: `teamcity-configs:generate` clean, **with a negative control** — a typo in the registered object name fails with `Compilation error settings.kts:… Unresolved reference`. (`mvn test` does **not** validate this DSL: it reports BUILD SUCCESS while logging "Nothing to compile".)
- The stats script matches PIT's own numbers on a real report; a missing report prints a WARNING and exits 0, so a path problem cannot turn a threshold pass red.

## Not verified yet

`[1.1]` cannot run before this merges — TeamCity loads versioned settings from the default branch, so the config does not exist on the server yet. Two behaviours are therefore consistent-by-construction but unobserved: the **artifact layout** and the **statistics step on a real agent**. Please run `[1.1]` manually right after merge; if the layout differs, the fix is one line.

## Review history

Sixteen findings, all accepted or answered:

| Finding | Resolution |
|---|---|
| P1 — write-scoped `GITHUB_TOKEN` on a job running PR-controlled build logic | `permissions: contents: read` |
| P2 — `paths` missed the tests that actually determine the score (243 mutants are killed by `..server.service.impl..`) | trigger on the whole server `src/test/**` |
| P2 — the summary counted `KILLED` while the threshold uses `detected` | count `detected='true'`, broken down by status |
| P2 — the statistics step was skipped exactly when PIT failed on the threshold | `ExecutionMode.ALWAYS` on that step |
| P2 — the unified index link was never regenerated or published in CI | a `qualityReportsIndex` step (also ALWAYS) + `build/reports/quality/**` |
| P2 — the artifact rule's static prefix flattened the path, so the index link 404'd | leading `**`, preserving the repo-relative structure |
| P2 — the documented artifact path did not match that rule | corrected to the real `reports/pitest/<module>/build/reports/pitest/index.html` |
| P2 — thresholds were called ratchets; the workflow was called a gate | floors, and a check — with the one change that would make it blocking named |
| P2 — measurement snapshots sat in source comments | moved to `quality-baseline.md`; the comments keep only the invariant |
| P3 — `settings.gradle` (where the plugin is declared) was not a trigger path | added |
| P3 — the new config was missing from `buildTypesOrder`, and its number reused a retired one | `[1.1]` with matching id `11MutationTestingAuto`, placed after `[1.0]` |
| P3 — `[MANUAL]` contradicted the weekly schedule | `[AUTO]`, with the reason recorded next to the trigger |
| P3 — the empty `dependencies {}` block guaranteed nothing | removed; the real inheritance limits documented instead |
| P3 — repeated pushes ran duplicate 8-minute analyses | concurrency group with `cancel-in-progress` |
| P3 — "the only check that fails an assertion-free test" overclaimed | restated around the asymmetry: a survivor proves nothing noticed; a kill only proves something did |
| P3 — mutator counts in the description mixed SURVIVED with all undetected | recomputed and split (342 survived vs 723 undetected) |

Closed with reasoning rather than a change: SHA-pinning the actions (the repository pins by tag everywhere — a single-file exception buys no safety), the parser shared with the TeamCity stats script (two different CI consumers; the duplication is now cross-referenced in both files so a change to one is a change to both), and a numbered SYS requirement (of 355 SYS entries none covers CI/build tooling — ArchUnit, the linters and the coverage floors all merged without one).

## Follow-ups

- Promote `mutation-test` into `gate-merge.needs` once the floor has held for a few weeks — that single change makes it blocking.
- Kill the mapper survivors **before** the planned property-based work on the version model: `VersionRangePartition` is already the strongest large class (74%), while the mapper layer leaves the wire path unchecked.
- Give unit tests to `ConfigurationRowAccessorsKt` and `MappersKt`, currently reachable only from `dbTest`.

## Summary by CodeRabbit

- **New Features**
  - Added mutation testing for server utility and mapper code to identify insufficient test coverage.
  - Added mutation score and coverage thresholds to track quality over time.
  - Added generated quality-report links and summaries for mutation testing results.

- **Quality & Reporting**
  - Added weekly automated mutation analysis with downloadable reports and trend statistics.
  - Reports distinguish detected, surviving, timed-out, and uncovered mutations.

- **Documentation**
  - Documented mutation testing scope, thresholds, interpretation, and non-blocking CI behavior.


pgorbachev:
## Review comments — disposition

Going through the CodeRabbit findings so a human reviewer does not have to re-derive which still apply. Five were fixed in later commits on this branch; three are closed with reasoning; one is fixed by this push.

**Fixed on the branch (verified in the current tree, not just intended):**

| Finding | Where it now stands |
|---|---|
| Add a concurrency group to cancel superseded runs | `mutation.yml:20-22` — `mutation-${{ github.workflow }}-${{ github.ref }}`, `cancel-in-progress: true`. Worth doing precisely because the run costs ~8 min: the first CI run measured **8m38s**. |
| "the per-change gate" contradicts non-gating | `non-functional-spec.md:360` now reads "per-change **check** (reporting, not …)". |
| Document the actual TeamCity artifact path | `quality-baseline.md:295` gives the full `reports/pitest/components-registry-service-server/build/reports/pitest/index.html`. |
| Move measurement snapshots out of source comments | Removed from `components-registry-service-server/build.gradle`; the comments keep only the invariant, the numbers live in the dated ledger. |
| Keep `docs/registry/` current-state | Addressed by this push — see below. |

**Closed with reasoning:**

- **Pin the actions to commit SHAs.** This repository pins by tag in every workflow, including the shared `octopus-base` ones. A single file pinning by SHA buys no real safety while the rest of the surface stays tag-pinned; it would need to be a repository-wide decision to mean anything. Also note the web query in that comment answered about `setup-java@v5`, not the `v4` in use here.
- **The counting is duplicated with `scripts/teamcity/report-mutation-stats.sh`.** Deliberate: two different CI consumers, a few greps each, and no shared runtime between a GitHub job summary and TeamCity service messages. The duplication is now cross-referenced in both files with an explicit instruction that a change to one belongs in the other in the same commit — which is exactly how the divergence found in this review round got fixed in both places at once.
- **Add a numbered SYS requirement.** Of the 355 SYS entries, none covers CI or build tooling; ArchUnit, the linters and the coverage floors all merged without one. Inventing an identifier here would create a trace that leads nowhere. If CI checks should be requirement-traced, that is a repository-wide convention change, not a condition on this PR.

**Fixed by this push — the `docs/registry/` current-state rule.** The finding is correct that `docs/registry/README.md` states the folder describes the system as it is today, with history in ADRs and `.github/audit/`. But `quality-baseline.md` cannot work that way: a measurement without its date and commit is not verifiable, and a "before / after" claim needs both readings side by side. The file already accumulated dated snapshots before this PR. Rather than leave the tension implicit, the header now names it: this file is the deliberate exception, dated entries are never rewritten, and a number in a row is a record of what a threshold was on that date — the enforced value lives in the build configuration and only there.

## One finding from an independent review round, since it is the substantive change here

The mutation score shown in the GitHub summary and charted in TeamCity was computed with shell integer division, which truncates, while the floor is compared against PIT's own number. Near the floor the reported value could sit a point below the one that actually decided pass or fail.

Verified rather than assumed, by decompiling the pinned engine: the threshold is checked against `MutationStatistics.getPercentageDetected()`, which is `PercentageCalculator.getPercentage(total, detected)` = `min(99, Math.round(100.0f / total * detected))` — the division happens first and in **single precision**, so it is not plain round-half-up either.

Both files now round half up and cap at 99. Exact parity is impossible in integer shell arithmetic, and the comments say so instead of claiming it: sweeping the real calculator over every count up to `total=20000` (200M pairs), this form deviates only on a score landing on an exact half — 903 cases, **every one of them a point low, none high** — where truncation was wrong on roughly half of all counts. A double-precision variant was checked too and deviates on exactly the same pairs, so moving to `awk` would buy nothing.

Also surfaced: mutations whose status falls outside the four charted buckets (`NON_VIABLE`, engine errors), so the numbers cannot quietly stop adding up to the total.

rodemark (approved):

PR #472, opened 2026-08-02
ci: gate PRs on spec-delta and spec-first

pgorbachev:
## What

`AGENTS.md` § Technical Requirements Tracking already requires "requirements first, then implementation", but nothing enforced it, so in practice requirements were written at the end of a change if at all. A requirement written after the code documents what got built rather than agreeing what to build.

Adds a blocking `spec / gate` job that makes that rule enforceable:

- **spec-delta** — a PR touching `*/src/main/{kotlin,java,resources}/**` or `components-registry-automation/data/**` must also touch a requirement, spec, ADR, `api-changelog.md`, or design doc.
- **spec-first** — the first spec commit must precede the first behaviour commit on the branch.

Both are skipped by applying the `no-spec-impact` label. Nothing in the PR body opts out.

## Scope decisions worth checking

- **`src/main/resources` counts as behaviour.** Flyway migrations change observable schema, and the `octopus-security` role-to-permission mapping — who may edit what — lives in `application.yml`. The `-test` and `-ft-db` profile configs and the generated `openapi/v4.json` are excluded.
- **`components-registry-automation/data/` counts too.** That module keeps its runtime configuration — supported groupIds and systems, version-name and product-type mappings — outside `src/main`. Its `okd/` stand template and `metarunners/` build steps do not count.
- **`api-changelog.md` satisfies the gate.** It is the document this repo instructs contributors to update on a v4 surface change; without it, doing exactly what the docs say still failed.

## Notes for review

The script is byte-identical with the Portal copy (octopusden/octopus-components-management-portal#213) — all repo-specific configuration is passed by environment from the workflow — and carries its own 42-case suite, run in the same job. One case asserts the regexes here match the ones in `merge-gate.yml`.

`.gitignore` gains `!.github/scripts/*.sh`: the repo ignores `*.sh` globally, so without it the scripts silently vanish from the commit and the workflow fails on a missing file.

## Verification

`bash .github/scripts/spec-gate.test.sh` → 42 passed, 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Added automated merge checks requiring behavior changes to include corresponding specification updates.
  * Enforced specification-first commit ordering before changes can be merged.
  * Added an approved opt-out for changes with no specification impact.
  * Checks now rerun when pull request labels or details change.

* **Documentation**
  * Added pull request guidance for describing changes, linking specifications, and documenting verification.
  * Documented specification requirements, exclusions, recognized files, and opt-out rules.

* **Tests**
  * Added comprehensive coverage for specification detection, commit ordering, exceptions, renames, merges, and unusual paths.

## Changed since the first push

Mirrors octopusden/octopus-components-management-portal#213 — the script stays byte-identical:

- **The opt-out is a label only**; the PR-body checkbox is gone after five review rounds turned up nine ways to fake it.
- **The gate fails closed** on an unresolvable base, a failed git call, or a path git had to quote — all three previously read as "nothing changed" and passed.
- **`labeled`/`unlabeled` added to the workflow triggers**, so removing the label cannot leave a stale green on the same SHA.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,24 @@
+## What
+
+<!-- One paragraph: what changes in service behaviour, and why. -->
+
+## Requirement / spec
+
+<!--
+This repo is spec-first (AGENTS.md § Technical Requirements Tracking rule 2):
+a behaviour change must update a numbered requirement or a spec document, and
+that spec must be committed *before* the implementation. The `spec / gate`
+check enforces both.
+
+Link the SYS-NNN / MIG-NNN / RES-NNN entry or spec this PR implements.
+-->
+
+Spec:
+
+- [ ] no-spec-impact — this PR changes no observable behaviour (refactor,
+      tests, build/CI, dependency bump, docs-only). Ticking this skips the
+      spec gate; leave it unticked otherwise.
+
+## Verification
+
+<!-- Commands run (./gradlew test, dbTest, qualityStatic), and manual checks. -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,9 +15,14 @@ Link the SYS-NNN / MIG-NNN / RES-NNN entry or spec this PR implements.
 
 Spec:
 
-- [ ] no-spec-impact — this PR changes no observable behaviour (refactor,
-      tests, build/CI, dependency bump, docs-only). Ticking this skips the
-      spec gate; leave it unticked otherwise.
+<!--
+No requirement needed? Apply the `no-spec-impact` label — that, and only that,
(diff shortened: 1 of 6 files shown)


fishinitself (approved):

pgorbachev:
Superseded by #493, which keeps the mechanical half only: a change to the committed v4 contract must carry an `api-changelog.md` entry. The broad spec-delta/spec-first rule is dropped — a path-based check cannot tell a feature from a bug fix, so it would have fired on every rename and migration and made `no-spec-impact` routine. Whether a change needs a requirement stays a review judgement, now stated as such in AGENTS.md.

